PR #41, opened 2025-09-15
docs: :memo: add requirement about CLI

lwjohnst86:
# Description

That `check-datapackage` can also be used as a CLI (potentially).

This PR needs a quick review.

## Checklist

- [x] Ran `just run-all`


martonvago (changes requested):
:+1: :+1:

signekb (approved):

PR #47, opened 2025-09-16
docs: :memo: add check and issue guide

### LINKED ISSUE #48: Draft of a simple guide for using `check-datapackage`
Can revise a lot of what is already there, but will need to be updated to match the design.

martonvago:
# Description

This PR adds a guide about how to use `check()`.

Closes #48 

Needs an in-depth review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`


docs/guide/check.qmd
```diff
@@ -0,0 +1,329 @@
...
+    defined in the standard.
+-   `strict`: Whether to run recommended checks in addition to required
+    ones. Defaults to `False`.
+
+### Excluding checks
+
+You can exclude checks based on their `type` and the fields they apply
+to.
+
+The Data Package standard defines a range of check types (e.g.,
```
martonvago, line 129: Technically the types are defined by JSON schema and Data Package uses a subset of these

docs/guide/check.qmd
```diff
@@ -0,0 +1,329 @@
...
+Python `dict` as input, representing the descriptor, and by default
+returns a list of `Issue` objects, each corresponding to a failed check
+on a field in the descriptor. If you prefer the output as an error, you
+can also run `check()` in error mode. When checking a descriptor, you
+can customise the function’s behaviour by specifying issues to ignore
+and additional custom checks to perform.
+
+## Default mode
+
+When you run `check()` in default mode, it returns a list of `Issue`
```
lwjohnst86, line 25:
```suggestion
When you run `check()`, by default it returns a list of `Issue`
```

docs/guide/check.qmd
```diff
@@ -0,0 +1,329 @@
...
+
+config = cdp.Config(rules=[license_rule])
+issues = cdp.check(descriptor=package_descriptor, config=config)
+print(issues)
+```
+
+We can see that the custom rule was applied: `check()` returned one
+issue flagging the first license attached to the Data Package.
+
+### Strict mode
```
lwjohnst86, line 262: This section could stay here in the `check` guide

docs/guide/check.qmd
```diff
@@ -0,0 +1,329 @@
...
+            "path": "resources/woolly-dormice-2015/data.parquet",
+        }
+    ],
+}
+
+issues = cdp.check(descriptor=package_descriptor, strict=True)
+print(issues)
+```
+
+## Understanding `Issue`s
```
lwjohnst86, line 296: This could be it's own page

docs/guide/check.qmd
```diff
@@ -0,0 +1,329 @@
...
+issues = cdp.check(descriptor=package_descriptor, config=config)
+print(issues)
+```
+
+We can see that the custom rule was applied: `check()` returned one
+issue flagging the first license attached to the Data Package.
+
+### Strict mode
+
+The Data Package standard expresses both required and recommended rules.
```
lwjohnst86, line 264:
```suggestion
The Data Package standard has both required and recommended rules.
```

docs/guide/check.qmd
```diff
@@ -0,0 +1,329 @@
...
+        }
+    ],
+}
+
+config = cdp.Config(exclude=[exclude_required, exclude_name, exclude_path])
+issues = cdp.check(descriptor=package_descriptor, config=config)
+print(issues)
+```
+
+In the example above, we would expect four checks to fail: the package
```
lwjohnst86, line 189:
```suggestion
In the example above, we would expect four `Issue` items: the package
```

docs/guide/check.qmd
```diff
@@ -0,0 +1,329 @@
...
+    "title": "Hibernation Physiology of the Woolly Dormouse: A Scoping Review.",
+    "id": "123-abc-123",
+    "created": "2014-05-14T05:00:01+00:00",
+    "version": "1.0.0",
+    "licenses": [{"name": "odc-pddl"}],
+}
+
+issues = cdp.check(descriptor=package_descriptor)
+print(cdp.explain(issues))
+```
```
lwjohnst86, line 329:
````suggestion
```

When using `check(error=True)`, the `explain()` function will be used by default as
the error messages.
````

**Review by lwjohnst86 — CHANGES_REQUESTED** [review]
Super nice!!!! :tada: :taco: Some suggested changes :relaxed:

File: docs/guide/check.qmd [hunk 8]
```diff
@@ -0,0 +1,329 @@
...
+the available properties at each level of the descriptor, specifies
+which properties are required, and describes the allowed values for
+each.
+
+`check-datapackage` provides the `check()` function to verify that a
+Data Package descriptor meets these requirements. `check()` takes a
+Python `dict` as input, representing the descriptor, and by default
+returns a list of `Issue` objects, each corresponding to a failed check
+on a field in the descriptor. If you prefer the output as an error, you
+can also run `check()` in error mode. When checking a descriptor, you
```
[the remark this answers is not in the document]
martonvago: This looks a bit off to me because it creates the false impression that `check(error=True)` is correct code as is. But of course you need the descriptor as well. Maybe `check(..., error=True)`? Or they can scroll down to the section about this.

docs/guide/check.qmd
martonvago, line 329:
```suggestion
```

When setting `error=True` in `check()`, error messages will be generated
by the `explain()` function.
```

File: docs/guide/check.qmd (hunk 8 above)
Comment (martonvago) [inline docs/guide/check.qmd:19]:
```suggestion
can also run `check()` with `error=True`. When checking a descriptor, you
```

martonvago:
I moved "Understanding issues" out, will move "Configuration" in another PR then

lwjohnst86 (approved):

PR #59, opened 2025-09-18
feat: :sparkles: add `Config`, `Rule`, `Exclude`

### LINKED ISSUE #24: Config items as dataclasses to ensure types, defaults, and available options?

### LINKED ISSUE #42: Refactor/create `Config` object
Might need to mostly create this from scratch, including the Rule and Exclude classes

martonvago:
# Description

This PR adds `Config`, `Rule`, and `Exclude`. For now, only the attributes are included.

Closes #42, closes #24

Needs an in-depth review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`

### DIFF
--- a/src/check_datapackage/config.py
+++ b/src/check_datapackage/config.py
@@ -1,0 +1,40 @@
+from dataclasses import dataclass, field
+from typing import Literal
+
+from check_datapackage.exclude import Exclude
+from check_datapackage.rule import Rule
+
+
+@dataclass
+class Config:
+    """Configuration for checking a Data Package descriptor.
+
+    Attributes:
+        exclude (list[Exclude]): Any issues matching any of these exclusions will be
+            ignored (i.e., removed from the output of the check function).
+        rules (list[Rule]): Custom checks listed here will be done in addition
+            to checks defined in the Data Package standard.
+        strict (bool): Whether to run recommended as well as required checks. If
+            True, recommended checks will also be run. Defaults to False.
+        version (str): The version of the Data Package standard to check against.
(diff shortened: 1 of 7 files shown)


src/check_datapackage/config.py
```diff
@@ -0,0 +1,40 @@
...
+            target="$.licenses[*].name",
+            message="Data Packages may only be licensed under MIT.",
+            check=lambda license_name: license_name == "mit",
+        )
+        config = cdp.Config(exclude=[exclude_required], rules=[license_rule])
+        ```
+    """
+
+    exclude: list[Exclude] = field(default_factory=list)
+    rules: list[Rule] = field(default_factory=list)
```
martonvago, line 38: This is needed instead of `[]` to make sure that a new list is created for each `Config` instance. With just `[]` as the default, all `Config` instances with the default would share the same list between them :upside_down_face:

src/check_datapackage/exclude.py
```diff
@@ -0,0 +1,34 @@
...
+
+@dataclass
+class Exclude:
+    """Exclude issues when checking a Data Package descriptor.
+
+    When both `target` and `type` are provided, an issue has to match both to be
+    excluded.
+
+    Attributes:
+        target (str | None): [JSON path](https://jg-rp.github.io/python-jsonpath/syntax/)
```
martonvago, line 12: Changed the link to an explanation of JSON path written by the package we'll use. This explains the syntax exactly as the package implements it.

lwjohnst86 (approved):
Nice! :grin:

PR #67, opened 2025-09-19
feat: :sparkles: add `Issue` class

### LINKED ISSUE #44: Create `Issue` class
Will most likely need to create this from scratch

### LINKED ISSUE #54: Map `ValidationError` to `Issue`
### Task
`jsonschema` returns `ValidationError`s. We have to map these to `Issue`s, like we map them to `CheckError`s currently.
#### Challenges: 
- Handling grouped `ValidationError`s. We have to list all possible cases from the Data Package JSON schema (not that many) and make sure they map to one `Issue` each. See #15 .
- Processing error messages. `ValidationError`s come with an error message but we want to swap this for a more detailed message. We have to decide if we want to do this when mapping to `Issue` or later in `explain()`.

### Output
A function like `validation_error_to_issue(validation_error: jsonschema.ValidationError) -> Issue`

martonvago:
# Description

This PR adds the `Issue` class.
I haven't deleted `CheckError` yet because it is used in the current excluding logic still.

Closes #44 closes #54

Needs an in-depth review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`

### DIFF
--- a/src/check_datapackage/issue.py
+++ b/src/check_datapackage/issue.py
@@ -1,0 +1,32 @@
+from dataclasses import dataclass
+
+
+@dataclass
+class Issue:
+    """An issue found while checking a Data Package descriptor.
+
+    One `Issue` object represents one failed check on one field within the descriptor.
+
+    Attributes:
+        location (string): A [JSON path](https://jg-rp.github.io/python-jsonpath/syntax/)
+            format pointing to the field in the input object where the issue is located.
+            For example, `$.resources[2].name`.
+        message (string): A description of what exactly the issue is.
+        type (string): The type of the check that failed. Used mostly for excluding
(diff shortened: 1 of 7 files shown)


src/check_datapackage/issue.py
```diff
@@ -0,0 +1,32 @@
+from dataclasses import dataclass
+
+
+@dataclass(order=True, frozen=True)
```
martonvago, line 4: This is to allow comparison of issues for things like sorting and calling `set` on them.

lwjohnst86 (changes requested):
Very nice! Just some small fixes

lwjohnst86 (approved):
:tada: Greaattttt :grin:

PR #74, opened 2025-09-22
feat: :sparkles: exclude by JSON schema type in `check()`

### LINKED ISSUE #51: Implement exclude logic
Given an `Exclude` object, a list of `Issue`s, and a descriptor, we want to be able to filter out the issues matched by the `Exclude` object.

This logic could be implemented as a standalone function (usage: `issues = exclude_issues(exclude, issues, descriptor)`) or as a method on `Exclude` (usage: `issues = exclude.exclude(issues, descriptor)`).

### LINKED ISSUE #71: Exclude by types
To keep the PR small, I'll start with excluding by type first. Then take on the targets afterwards.

lwjohnst86:
# Description

This implements the exclusion, by type only right now. The target exclusion seems to be a bit more complicated, so I wanted to split PR into first the type, then the target later.

Closes #71, closes #51

Needs an in-depth review.

## Checklist

- [x] Ran `just run-all`


martonvago (changes requested):
Niiiiiice :star2:

martonvago (changes requested):
Just minor things ☺️

martonvago (approved):

Issue #15, opened 2025-01-14
Improve the handling of complex, summary-type `jsonschema` errors


martonvago:
Currently, when we transform  `jsonschema`'s `ValidationErrors` into `CheckErrors`, we don't preserve relationships between errors. For example, in the case where any of two fields must be present on an object, if both fields are missing, we output a `CheckError` for both. From this output alone, it is not possible to tell that filling in either of those fields would have resolved the issue.

One idea is to add a `group` attribute to `CheckError` that would make it possible to retain these relationships between errors. So in a case like the above, the output would be:
```json
[
    {
        "message": "'data' is a required property",
        "json_path": "$.data",
        "validator": "required",
        "group": "oneOf-1",
    },
    {
        "message": "'path' is a required property",
        "json_path": "$.path",
        "validator": "required",
        "group": "oneOf-1",
    },
]
```
Using this output, it is possible to tell which errors belong to one group, and the name of the group implies that only one of the errors needs to be fixed.

Tasks:
- [ ] Add a `group` attribute to `CheckError`. Should be optional.
- [ ] Modify `validation_errors_to_check_errors` to populate the `group` field of `CheckErrors`. For this, an initial strategy might be looking at each `ValidationError` in the flattened list and checking if it is a child of a complex / summary-type error (validators for this listed in `COMPLEX_VALIDATORS`). `ValidationError`s should have this information on them. When naming the group, keep the name unique within the list of errors (there may be multiple `oneOf`-type groups!) -- e.g., add a number.
- [ ] Update tests.
- [ ] Consider whether the information captured by `group` should be translated into a more user-friendly error message when the errors are displayed. For this, look at `FailedCheckError` and how it displays its list of `CheckError`s.

martonvago:
I've had another look at these grouped errors and just wanted to say that maybe there are simpler options than what I suggested here. 

Basically, we're only looking at `anyOf` and `oneOf` rules in the JSON schema, and these don't come up thaaat frequently. And when they do come up, the error structures under them are very similar.

So we could
- Detect grouped `ValidationError`s in the `jsonschema` output and map them to one `CheckError` with a custom message. E.g. "One of `data` or `path` must be provided." We can do this because we can anticipate these grouped errors based on the JSON schema.
- Filter out `ValidationError`s that we know are irrelevant.
[...]

PR #81, opened 2025-09-23
refactor: :recycle: simplify handling of grouped errors

martonvago:
# Description

This PR refactors how the grouped errors listed in #15 could be simplified.
It adds custom logic for the first 2 items in the [list](https://github.com/seedcase-project/check-datapackage/issues/15#issuecomment-3320059951).

We will probably be able to share quite a bit of code between the various error cases, this is just a start.

Needs an in-depth review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`


lwjohnst86 (changes requested):
Nice! Question: Is the title "demo" mean anything? Like, you don't want this to be merged in or?

src/check_datapackage/internals.py
```diff
@@ -100,46 +99,110 @@ def _validation_errors_to_issues(
...
+            "path",
+            "oneOf",
+        ],
+    ):
+        return _handle_S_resources_x_path(sub_errors)
+
+    return [_create_issue(sub_error) for sub_error in sub_errors]
+
+
+def _handle_S_resources_x(sub_errors: list[ValidationError]) -> list[Issue]:
```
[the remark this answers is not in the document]
martonvago: It wants to express `# Handle issues at $.resources[x]`, maybe more obvious where it's called. We can call these functions something else, but the names should be systematic and express precisely which validation errors are handled by each function.

src/check_datapackage/exclude.py
martonvago, line 1: If these are reused, then makes sense to put them in `internals`

src/check_datapackage/internals.py
```diff
@@ -106,11 +83,7 @@ def _validation_errors_to_issues(
         A list of `Issue`s.
     """
     return sorted(
-        {
-            issue
-            for error in validation_errors
-            for issue in _validation_error_to_issues(error)
-        }
+        set(chain.from_iterable(map(_validation_error_to_issues, validation_errors)))
```
martonvago, line 86: `chain.from_iterable` flattens a list of lists

src/check_datapackage/internals.py
```diff
@@ -137,30 +110,27 @@ def _validation_error_to_issues(error: ValidationError) -> list[Issue]:
...
     """Do not flag missing `path` and `data` separately."""
-    issues: list[Issue] = []
-    path_or_data_required_error: ValidationError | None = None
-
-    for error in sub_errors:
-        path = _get_full_json_path_from_error(error)
+    path_or_data_required_errors, other_errors = _partition(
+        sub_errors,
+        lambda error: str(error.validator) == "required"
+        and _get_full_json_path_from_error(error).endswith(("path", "data")),
```
martonvago, line 121: Can extract the lambda into a separate function if you want

martonvago:
> Nice! Question: Is the title "demo" mean anything? Like, you don't want this to be merged in or?

We can merge it in if you like this approach! I just wasn't sure if anyone had a better idea for dealing with all the custom mapping. There will be 15 cases all in all, so that will require lots of rather specific and arcane code.

src/check_datapackage/internals.py
```diff
@@ -76,46 +76,103 @@ def _validation_errors_to_issues(
...
     Returns:
         A list of `Issue`s.
     """
-    issues = [
+    return sorted(
+        set(chain.from_iterable(map(_validation_error_to_issues, validation_errors)))
+    )
+
+
+def _validation_error_to_issues(error: ValidationError) -> list[Issue]:
```
lwjohnst86, line 90: If these are only used in one location, they can stay in that file. I think how you had it previous was fine ☺️

src/check_datapackage/internals.py
```diff
@@ -76,46 +76,103 @@ def _validation_errors_to_issues(
...
+    """Maps a `ValidationError` to one or more `Issue`s."""
+    if not error.context:
+        return [_create_issue(error)]
```
lwjohnst86, line 93: Not completely sure what this means or does. Is this if there are no errors? or only one? If it's only one, maybe you move this up into the function above so that it gets returned right there and can bypass the `chain` and `sorted`.

lwjohnst86 (changes requested):
Nice progress! Some more feedback ☺️

src/check_datapackage/internals.py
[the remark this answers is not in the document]
martonvago: This is the function that will handle all 15 cases though, so many more returns are coming. Do you have a more concrete view of what it should look like?

src/check_datapackage/internals.py
martonvago: This is when an error is a simple error without any sub-errors, not a grouped one. Basically, if we only had these errors, then we wouldn't need to have this PR. There can be many simple errors.

If you want I can handle simple and grouped errors separately by doing something like:
```python
simple_errors, grouped_errors = _partition(
    list(validation_errors), lambda error: not error.context
)
simple_issues = _map(simple_errors, _create_issue)
grouped_issues = chain.from_iterable(
    _map(grouped_errors, _validation_error_to_issues)
)

all_issues = chain(simple_issues, grouped_issues)
return sorted(set(all_issues))
```
in `_validation_errors_to_issues` above. Not sure if this is better than how it is now 🤷 .

src/check_datapackage/internals.py
```diff
@@ -76,46 +76,103 @@ def _validation_errors_to_issues(
...
+                ),
+                location=path_or_data_required_errors[0].json_path,
+                type="required",
+            )
+        )
+
+    return issues
+
+
+def _handle_S_resources_x_path(sub_errors: list[ValidationError]) -> list[Issue]:
```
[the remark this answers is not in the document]
martonvago: It returns one or more issues (see the `_map`)

src/check_datapackage/internals.py
```diff
@@ -134,3 +191,25 @@ def _get_full_json_path_from_error(error: ValidationError) -> str:
...
+
+
+def _partition(
+    items: Iterable[In], condition: Callable[[In], bool]
+) -> tuple[list[In], list[In]]:
+    """Split items into two lists based on a condition."""
+    return (
+        _filter(items, condition),
+        _filter(items, lambda item: not condition(item)),
+    )
```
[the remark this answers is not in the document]
martonvago: The `TypeVar` is for `_map` and `_filter` too (same considerations).

We can lose `_partition` for sure haha -- I was actually thinking you'd like it because it's a common method in languages that have a stream API, so smells functional.

lwjohnst86 (changes requested):

src/check_datapackage/internals.py
```diff
@@ -76,46 +76,108 @@ def _validation_errors_to_issues(
...
-      - sorted by error location
-
     Args:
         validation_errors: The `jsonschema.ValidationError`s to transform.
 
     Returns:
         A list of `Issue`s.
     """
-    issues = [
+    issue_groups = map(_validation_error_to_issues, validation_errors)
```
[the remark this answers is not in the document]
martonvago: So, e.g., have an object like:
```python
@dataclass
class JsonSchemaError:
    # Just copied the relevant fields of ValidationError as a draft
    message: str
    validator: str
    json_path: str
    absolute_path: Sequence[str | int]
    schema_path: deque[str | int]
    context: list["JsonSchemaError"] | None
```
and use this in the logic that sorts out grouped errors (under `context`)?

I can for sure see how we could simplify some things (e.g. wouldn't need to keep casting `validator` to `str`), but wouldn't the logic for handling `context` errors be exactly the same?

Or if you meant something else, I would be very happy to look at your draft!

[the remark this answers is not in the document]
lwjohnst86: Yea! Except, we don't actually need `context`. We need the stuff inside it, but not it itself. So it doesn't need to be in this type of class. And you could call it something simpler, like `SchemaError`.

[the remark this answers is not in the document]
martonvago: Maybe I'm misunderstanding you, but what we are doing here is getting rid of the grouping. `jsonschema` gives us some grouped errors and we are ungrouping them.

Here I'm doing that by processing each group separately and deciding what (simple, ungrouped) issues the errors in the group correspond to.

Are you suggesting to ungroup the errors first and then process them? I can have a look if enough information is retained on ungrouped errors to map them to issues. This was the original approach by the way (but we only dealt with a subset of cases).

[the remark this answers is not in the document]
lwjohnst86: Yea, extract all relevant information, and then do filtering of that. Rather than digging down, filtering, and extracting at the end.

src/check_datapackage/internals.py
```diff
@@ -68,11 +74,31 @@ def _check_object_against_json_schema(
...
+            Path components are separated by '/'.
+        json_path (str): The JSON path to the field that violates the rule.
+        parent (Optional[SchemaError]): The error group the error belongs to, if any.
+    """
+
+    message: str
+    validator: str
+    schema_path: str
+    json_path: str
+    parent: Optional["SchemaError"] = None
```
martonvago, line 98: Instead of the children (the `context`), I kept the parent. This is useful because when we process the group, it allows us to consider only errors in the group -- this feels less error prone to me

src/check_datapackage/internals.py
```diff
@@ -82,102 +109,114 @@ def _validation_errors_to_issues(
...
     # Handle issues at $.resources[x]
-    if _schema_path_ends_in(error, ["resources", "items", "oneOf"]):
-        return _handle_S_resources_x(sub_errors)
+
+    resources_x_oneOf = _next(
+        schema_errors,
+        lambda error: error.schema_path.endswith("resources/items/oneOf"),
+    )
+    if resources_x_oneOf:
+        schema_errors = _handle_S_resources_x(resources_x_oneOf, schema_errors)
```
martonvago, line 121: 1. Find the parent error
2. If it exists, handle all child/group errors

src/check_datapackage/internals.py
```diff
@@ -198,13 +237,44 @@ def _get_full_json_path_from_error(error: ValidationError) -> str:
...
-    return list(filter(fn, x))
+    Returns:
+        A flat list of `SchemaError`s.
+    """
+    unwrapped = []
+    for error in errors:
+        unwrapped.append(_create_schema_error(error))
+        if error.context:
+            unwrapped.extend(_create_schema_errors(error.context))
+    return unwrapped
```
martonvago, line 254: This is the old function we had originally.
Not sure if you like the style, it's a pretty normal way to express recursion, but let me know if you have ideas for another way.

I'm using recursion because `error.context` may have multiple levels of nesting.

src/check_datapackage/internals.py
```diff
@@ -198,13 +237,44 @@ def _get_full_json_path_from_error(error: ValidationError) -> str:
...
 
-def _map(x: Iterable[In], fn: Callable[[In], Out]) -> list[Out]:
-    return list(map(fn, x))
+def _create_schema_error(error: ValidationError) -> SchemaError:
+    return SchemaError(
+        message=error.message,
+        validator=str(error.validator),
+        json_path=_get_full_json_path_from_error(error),
+        schema_path="/".join(_map(error.absolute_schema_path, str)),
+        parent=_create_schema_error(error.parent) if error.parent else None,  # type: ignore[arg-type]
```
martonvago, line 263: `# type: ignore[arg-type]` because `parent` is typed as `_Error` in `jsonschema` (the parent class of `ValidationError`), but this is not exposed, so can't use it. It's safe to view parent errors as `ValidationError`s.

tests/test_check_custom_error_mapping.py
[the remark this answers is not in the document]
martonvago: Will put it back eventually 👍 👍

src/check_datapackage/internals.py
```diff
@@ -71,6 +78,26 @@ def _check_object_against_json_schema(
     return _validation_errors_to_issues(validator.iter_errors(json_object))
 
 
+@dataclass
+class SchemaError:
```
martonvago, line 82: Called it this as discussed, but just to note, `jsonschema` also has a `SchemaError` and that is for when the schema itself is wrong. Unlikely to matter for us?

src/check_datapackage/internals.py
lwjohnst86: That's still using a for loop though. A for loop is still imperative, not declarative.

src/check_datapackage/internals.py
```diff
@@ -71,51 +79,150 @@ def _check_object_against_json_schema(
...
-        unwrapped.append(error)
-        if error.context:
-            unwrapped.extend(_unwrap_errors(error.context))
-    return unwrapped
+    return next(filter(condition, items), None)
+
+
+def flat_map(fn: Callable[[In], Iterable[Out]], items: Iterable[In]) -> Iterable[Out]:
+    """Maps and flattens the items by one level."""
+    return chain.from_iterable(map(fn, items))
```
martonvago, line 225: Returning `Iterable` felt nicer for my use case, so I called this `flat_map` without `_` and chose this arg order in order to align with the built-in `map` and `filter`. 

We could have it return a list though!

martonvago: Then I'll make it return a list instead and then (with the other changes) it will be properly aligned with our functionals :relaxed:

src/check_datapackage/internals.py
```diff
@@ -134,3 +241,43 @@ def _get_full_json_path_from_error(error: ValidationError) -> str:
...
+    if not error.context:
+        return current
+    return chain(
+        current,
+        flat_map(_validation_error_to_schema_errors, error.context),
+    )
+
+
+def _create_schema_errors(errors: list[ValidationError]) -> list[SchemaError]:
+    return list(flat_map(_validation_error_to_schema_errors, errors))
```
lwjohnst86, line 257:
```suggestion
    return flat_map(_validation_error_to_schema_errors, errors)
```

flat map should already output a list.

src/check_datapackage/internals.py
lwjohnst86, line 225:
```suggestion
    return list(chain.from_iterable(map(fn, items)))
```

src/check_datapackage/internals.py
```diff
@@ -134,3 +241,43 @@ def _get_full_json_path_from_error(error: ValidationError) -> str:
...
+
+
+def _is_path_type_error(error: SchemaError) -> bool:
+    return error.validator == "type" and error.json_path.endswith("path")
+
+
+def _create_issue(error: SchemaError) -> Issue:
+    return Issue(
+        message=error.message,
+        location=error.json_path,
```
lwjohnst86, line 281:
```suggestion
        json_path=error.json_path,
```

We've updated this to `json_path`. You'll have to update from main to get those changes.

lwjohnst86 (changes requested):
Great progress!! 🎉 some more comments ☺️

src/check_datapackage/internals.py
```diff
@@ -71,51 +79,150 @@ def _check_object_against_json_schema(
...
+        message (str): The error message generated by `jsonschema`.
+        validator (str): The type of the error.
+        schema_path (str): The path to the violated rule in the JSON schema.
+            Path components are separated by '/'.
+        json_path (str): The JSON path to the field that violates the rule.
+        parent (Optional[SchemaError]): The error group the error belongs to, if any.
+    """
+
+    message: str
+    validator: str
```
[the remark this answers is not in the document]
martonvago: Sure, we can align with `Issue`! I was aligning with `ValidationError`

src/check_datapackage/internals.py
```diff
@@ -48,88 +49,10 @@ def _add_resource_recommendations(schema: dict[str, Any]) -> dict[str, Any]:
...
-        The full `json_path` of the error.
-    """
-    if str(error.validator) == "required":
-        match = re.search("'(.*)' is a required property", error.message)
-        if match:
-            return f"{error.json_path}.{match.group(1)}"
-    return error.json_path
+def _flat_map(items: Iterable[In], fn: Callable[[In], Iterable[Out]]) -> list[Out]:
+    """Maps and flattens the items by one level."""
+    return list(chain.from_iterable(map(fn, items)))
```
martonvago, line 58: I think it makes sense to leave the functionals in internals

src/check_datapackage/check.py
```diff
@@ -40,3 +45,203 @@ class for more details, especially about the default values.
...
+
+    Args:
+        validation_errors: The `jsonschema.ValidationError`s to transform.
+
+    Returns:
+        A list of `Issue`s.
+    """
+    schema_errors = _flat_map(validation_errors, _validation_error_to_schema_errors)
+    grouped_errors = _filter(schema_errors, lambda error: error.type in GROUP_ERRORS)
+    schema_errors = reduce(_handle_grouped_error, grouped_errors, schema_errors)
```
martonvago, line 106: This is the equivalent of 
```python
for error in grouped_errors:
    schema_errors = _handle_grouped_error(schema_errors, error)
```
So we take each grouped error (well, the parent of the group) in turn and remove or add schema errors as needed.

Let me know if you like `reduce`. I've always thought it wasn't very intuitive 😅

src/check_datapackage/check.py
```diff
@@ -40,3 +45,203 @@ class for more details, especially about the default values.
...
```
martonvago, line 106: This is the equivalent of 
```python
for error in grouped_errors:
    schema_errors = _handle_grouped_error(schema_errors, error)
```
So we take each grouped error (well, the parent of the group) in turn and remove or add schema errors as needed.

Let me know if you like `reduce`. I've always thought it wasn't very intuitive 😅

lwjohnst86 (changes requested):
Awesome!! Just some very minor comments

lwjohnst86 (approved):
:tada: Amazing :tada:

PR #118, opened 2025-10-13
docs: :memo: update architecture naming section

### LINKED ISSUE #104: [discussion]: `Descriptor` vs `properties`
# Description

I think we need to be clearer on what we mean when we say *descriptor* and *property*/*properties*. I must say that I also find the distinction from the DataPackage standard confusing. Especially the fact that there are four descriptors and the rest are properties (`DataPackage`, `DataResource`, `TableDialect`, and `TableSchema`, respectively). I guess they have "chosen" these four to be descriptors bc they are the most essential ones in terms of describing the data itself? 

Maybe you understand this distinction better, @martonvago ? 
I've actually also asked about this distinction on the Frictionless Data Slack now.

Currently, our design docs include this definition: 

| Object | Description |
| -- | -- |
| descriptor | The standalone and complete metadata structure contained in a JSON file, for example, in `datapackage.json`. |
| properties | Individual elements of the metadata structure **or the structure as a whole.** |

For me, it becomes unclear/the definitions bleed into each other, with the last part of the properties description "or the structure as a whole".  

One way we *could* simplify this (without also having to rename anything in Sprout to keep consistency) would be to say that we only refer to descriptors as the content of the `datapackage.json` file.

That definition could look something like this:

| Object | Description |
| -- | -- | 
| descriptor | A standalone and complete metadata structure (a JSON object) contained in a JSON file, such as `datapackage.json`, or loaded as a Python dictionary |
| property/properties | Individual attribute(s) or metadata field(s), name-value pairs, of a descriptor, including nested properties. |

This way, whenever we load everything from the `datapackage.json` file in a `dict` it's the `descriptor`, while every subpart of it is referred to as properties. This is, however, not completely aligned with the DataPackage specification.

Another alternative would be to completely ignore the term "descriptor" and use properties for everything, as we've done so far. 

What do you think?

signekb:
# Description

This updates the naming section in the design docs.

Closes #104 
Related to #46 

Needs an in-depth review.

## Checklist

- [X] Formatted Markdown
- [X] Ran `just run-all`

### DIFF
--- a/docs/design/architecture.qmd
+++ b/docs/design/architecture.qmd
@@ -90,24 +90,20 @@ flowchart LR
 This section contains a naming scheme for `check-datapackage` that is
 inspired by the [Data Package standard](https://datapackage.org/).
 
-Following Data Package terminology, we simplify "data package" to
-"package" and "data resource" to "resource". The Data Package standard
-calls a standalone and complete metadata structure
-["descriptor"](https://datapackage.org/standard/glossary/#descriptor).
-Descriptors are JSON objects contained in JSON files (e.g.,
-`datapackage.json` contains the Data Package descriptor). We adopt this
-usage, but may also use the word "properties" to refer to the metadata
-as a whole.
+Overall, we follow the Data Package terminology where possible to keep
+things consistent. However, we also introduce some new terms and
+concepts specific to `check-datapackage`. The main objects and actions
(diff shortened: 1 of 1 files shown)


lwjohnst86 (changes requested):

lwjohnst86 (approved):
:tada:

PR #135, opened 2025-10-16
refactor: :truck: rename data package schema to include version

signekb:
# Description

Bc it's already in a `schemas/` folder, we don't need to include `schema` in the name. Instead, I've added the specific version (2.0) bc Data Package (hopefully?) will release a 2.1 at some point.

Needs a quick review.

## Checklist

- [X] Ran `just run-all`


lwjohnst86 (approved):
:+1: :grin:

PR #138, opened 2025-10-16
feat: :sparkles: exclude required issues at a given JSON path

### LINKED ISSUE #121: Excluding required rules

martonvago:
# Description

This PR adds some logic to handle excluding required issues at a given JSON path.

The main problem here comes down to the following:
Given `Exclude(type="required", jsonpath="$.resources[*].name")` and `Issue(type="required", jsonpath="$.resources[3].name")`, how can I detect that `$.resources[*].name` matches `$.resources[3].name`? I cannot get all matches in the descriptor for `$.resources[*].name` because, when the field is missing, there will be none.

One approach would be to decide it based on JSON path syntax. This will get thorny quite quickly and would be a partial reimplementation of `jsonpath` anyway.

Here, I went with the approach of taking `$.resources[3].name` and building an object that has this field: `{"resources": [{}, {}, {}, {"name": {}}]}`. It doesn't matter what the field contains, as long as it exists. Testing `$.resources[*].name` on this object (using `jsonpath`) works nicely.

I initially thought this was very hacky, but I'm coming around to it. But if anyone has better ideas, don't hold back!

Closes #121 

Needs an in-depth review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`


src/check_datapackage/internals.py
```diff
@@ -26,6 +26,12 @@ def _get_fields_at_jsonpath(
     return _map(matches, _create_descriptor_field)
 
 
+def _get_direct_jsonpaths(jsonpath: str, descriptor: dict[str, Any]) -> list[str]:
```
lwjohnst86: Should this `descriptor` argument be renamed then if it is being used for things that aren't `descriptor`s?

src/check_datapackage/exclude.py
```diff
@@ -70,19 +71,51 @@ def _get_matches(issue: Issue, exclude: Exclude, descriptor: dict[str, Any]) ->
...
+
+def _get_test_object_from_jsonpath(jsonpath: str) -> dict[str, Any]:
+    """Builds an object with a property at the given JSON Path location."""
+    fields = jsonpath.removeprefix("$.").split(".")
+    test_object: dict[str, Any] = {}
+    reduce(_set_object_field, fields, test_object)
+    return test_object
+
+
+def _set_object_field(obj: dict[str, Any], field: str) -> dict[str, Any]:
```
lwjohnst86: To make the purpose clearer here, maybe rename to `_set_field_value_to_placeholder()`?

lwjohnst86 (changes requested):
Just some initial comments and questions before doing a deeper dive :relaxed:

src/check_datapackage/exclude.py
```diff
@@ -70,19 +71,51 @@ def _get_matches(issue: Issue, exclude: Exclude, descriptor: dict[str, Any]) ->
...
         matches.append(_same_type(issue, exclude.type))
 
     return all(matches)
 
 
-def _same_jsonpath(issue: Issue, jsonpath: str, descriptor: dict[Any, str]) -> bool:
-    fields: list[DescriptorField] = _get_fields_at_jsonpath(jsonpath, descriptor)
-    jsonpaths: list[str] = _map(fields, lambda field: field.jsonpath)
+def _jsonpaths_match(issue: Issue, jsonpath: str) -> bool:
+    test_object = _get_test_object_from_jsonpath(issue.jsonpath)
```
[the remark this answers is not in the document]
martonvago: Sure! I just meant that it is an object to test JSON paths with

lwjohnst86 (changes requested):
Hmm, ok, there's more confusion then :thinking: :face_with_spiral_eyes:

lwjohnst86 (changes requested):
Great progress!! :tada: :tada: some minor edits and comments, otherwise :+1:

lwjohnst86 (dismissed):
Awesome :relaxed: just need to resolve merge conflicts, then :+1:

lwjohnst86 (approved):
Nice :+1:

PR #141, opened 2025-10-16
docs: :memo: simplify by removing mentions of "descriptor" from landing page

signekb:
# Description

This PR simplifies (in my opinion) the wording on the landing page to avoid the word "descriptor" and use either "metadata" or "`datapackage.json`" instead. 

Needs an in-depth review.

## Checklist

- [X] Formatted Markdown
- [X] Ran `just run-all`


lwjohnst86 (approved):
Very lovely :100:

PR #142, opened 2025-10-17
refactor: :recycle: rename `check` to `check_value` in `CustomCheck`

martonvago:
# Description

I thought there were so many "check"s around, and this method actually checks the value of a field, not the descriptor/properties. 

Needs a quick review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`

### DIFF
--- a/docs/guide/config.qmd
+++ b/docs/guide/config.qmd
@@ -103,7 +103,7 @@ license_check = cdp.CustomCheck(
         Data Packages may only be licensed under MIT. Please review
         the licenses listed in the Data Package.
         """),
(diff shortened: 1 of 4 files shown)


lwjohnst86 (approved):
This works fine with me. Though, what do you think of something like `condition` instead of `check_value`?

signekb (commented):
I actually don't mind it being "check" here, bc it's the actual check that the custom check will perform, right? :)

martonvago:
> This works fine with me. Though, what do you think of something like `condition` instead of `check_value`?

I like it!

PR #143, opened 2025-10-17
refactor: ♻️ rename `check()` arg from `descriptor` to `properties`

signekb:
# Description

Again following that we've agreed to use the word "properties" about the descriptor's content loaded as a `dict`. 

Related to #104 

Needs an in-depth review.

## Checklist

- [X] Formatted Markdown
- [X] Ran `just run-all`


docs/guide/check.qmd
```diff
@@ -1,36 +1,43 @@
 ---
-title: "Checking a Data Package descriptor"
+title: "Checking a Data Package's properties"
```
signekb: We could also use "metadata" here, as I've done on the landing page now #141

signekb: I've started to wonder if we can stop using the word "descriptor" entirely, using either metadata or `datapackage.json` instead (whichever is appropriate in the context). What do you think?

lwjohnst86: I'm ok with that :+1: simpler, clearer, plainer language is something I always appreciate!! :grin:

lwjohnst86 (approved):
:+1: looks good to me :grin:

PR #145, opened 2025-10-20
refactor: 🚚 rename `Exclude` class to `Exclusion`

### LINKED ISSUE #103: [discussion]: Naming of `Exclude` and `Rule`
# Description

In `Config`, we currently have `Exclude` and `Rule`(s), but I think we can find some better/clearer/more intuitive names for them, especially for `Rule` :seedling:

They are described as so: 
>@dataclass
>class Config:
    """Configuration for checking a Data Package descriptor.
>
>    Attributes:
        exclude (list[Exclude]): Any issues matching any of these exclusions will be
            ignored (i.e., removed from the output of the check function).
        rules (list[Rule]): Custom checks listed here will be done in addition
            to checks defined in the Data Package standard.

And in the classes themselves: 
>@dataclass
class Exclude:
    """Exclude issues when checking a Data Package descriptor.

>@dataclass
class Rule:
    """A custom check to be done on a Data Package descriptor.

So, as far as I understand, `Exclude` actually defines which checks to "ignore" or exclude from (the output of) the checks that are already in the schema, while `Rule` defines a custom check to be done, a check that is *added*

So, I think there might be something here with the words *ignore*, *exclude*, *custom*, and *add*. 
Could we rename `Exclude` and `Rule` to something like in the table below?

|`Exclude` | `Rule`| 
| --- | --- | 
| `ExcludeCheck` |`CustomCheck` |
| `Ignore` | `Add` |
| `Exclude` | `Custom` |

As we all know, naming is difficult, so let's discuss! What do you think?

signekb:
# Description

This includes updating docstrings, arguments, and text to match this change.

Closes #103 

Needs an in-depth review.

## Checklist

- [X] Formatted Markdown
- [X] Ran `just run-all`


lwjohnst86 (approved):
:tada: :truck:

PR #148, opened 2025-10-21
docs: :memo: align docstrings describing `type` arguments

### LINKED ISSUE #139: [discussion]: Change argument `type` in `CustomCheck`?
# Description 

Currently, we have multiple classes with an argument called `type`. However, what "type" means seems to differ across them (see overview below). Briefly, `CustomCheck` seems to stick out as it uses `type` as an identifier, not a "type" as  "required", "type", "pattern", or "format". So, could we change the name of that arg? Maybe to `name"? 

Or does it have to be `type` to fit with JSON Schema somehow? 

## Overview

In `CustomCheck`: 
```
class CustomCheck:
    ...
    type (str): An identifier for the custom check. It will be shown in error
            messages and can be used to exclude the check. Each custom check
            should have a unique `type`.
```
In the docstring example, it's `type="only-mit"`.

In `Exclude`:
```
class Exclude: 
    ...
    type (Optional[str]): The type of the issue to ignore (e.g., "required", "type",
            "pattern", or "format").
```
In the docstring example, it's `type="required"`.

In `Issue`: 
```
class Issue:
    ...
    type (string): The type of the check that failed. Used mostly for excluding
            specific types of issues.
```
In the docstring example, it's `type="required"`.

In `SchemaError`:
```
class SchemaError:
    ...
    type (str): The type of the error.
```
No docstring example. But in `check.py` in `_handle_S_resources_x()` and `_handle_S_resources_x_path`, it's `type="required` and `type="type"`, respectively.

signekb:
# Description

Based on the consensus in #139 that the docstring of `CustomCheck` needed to be updated, I've decided update and align the docstrings for all classes with the `type` argument. 

Closes #139 

Needs a quick review.

## Checklist

- [X] Ran `just run-all`


lwjohnst86 (approved):
:taco: nice alignment :+1:

PR #151, opened 2025-10-21
docs: :memo: update README with description of package

### LINKED ISSUE #65: Add description of package to README

lwjohnst86:
# Description

Closes #65

Needs a quick review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`


signekb (changes requested):
Nice with an update to this 👌 Some suggestions:

And: Would love if we could have insert the same text here and on the landing page, so we don't have to keep updating both 💭

signekb (changes requested):
Just one more suggestion, format the file, and then 👌 ✅

signekb (approved):

PR #158, opened 2025-10-27
refactor: :truck: rename to `extensions.py` from `custom_checks.py`

lwjohnst86:
# Description

As per #156, renaming custom checks to be extensions instead. Will implement the Extensions class in another PR.

Needs a quick review.

## Checklist

- [x] Ran `just run-all`


signekb (changes requested):
Nice---should this be singular though? So `extension.py` and `Extension` rather than `extensions.py` and `Extensions`? That would mirror the singular of `exclusion.py` and `Exclusion`.

martonvago (dismissed):

martonvago (dismissed):

PR #160, opened 2025-10-27
docs: :lipstick: use more sections in the reference sidebar

lwjohnst86:
# Description

Mostly aesthetic changes, so that there are multiple sections in the sidebar.

Needs a quick review.

## Checklist

- [x] Ran `just run-all`

PR #161, opened 2025-10-27
docs: :memo: add docstring examples to example properties

signekb:
# Description

When I reviewed #160 I noticed that there weren't examples for the example properties. So, I've added them here.

Needs a quick review.


lwjohnst86 (approved):
:tada:

PR #165, opened 2025-10-27
feat: :sparkles: adds `Extensions` meta-class

lwjohnst86:
# Description

Implements the `Extensions` meta-class. Also did some mypy/type hint fixes to the tests.

Needs an in-depth review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`

### DIFF
--- a/src/check_datapackage/config.py
+++ b/src/check_datapackage/config.py
@@ -12,7 +12,7 @@ class Config:
     Attributes:
         exclusions (list[Exclusion]): Any issues matching any of Exclusion objects will
             be excluded (i.e., removed from the output of the check function).
-        extensions (list[Extensions]): Additional checks (called extensions)
+        extensions (Extensions): Additional checks (called extensions)
             that supplement those specified by the Data Package standard.
         strict (bool): Whether to include "SHOULD" checks in addition to "MUST" checks
             from the Data Package standard. If True, "SHOULD" checks will also be
--- a/src/check_datapackage/config.py
+++ b/src/check_datapackage/config.py
@@ -1,7 +1,7 @@
 from dataclasses import dataclass, field
 from typing import Literal
 
-from check_datapackage.custom_check import CustomCheck
+from check_datapackage.custom_check import Extensions
 from check_datapackage.exclusion import Exclusion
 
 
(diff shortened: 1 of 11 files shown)


src/check_datapackage/config.py
```diff
@@ -12,8 +12,8 @@ class Config:
     Attributes:
         exclusions (list[Exclusion]): Any issues matching any of Exclusion objects will
             be excluded (i.e., removed from the output of the check function).
-        custom_checks (list[CustomCheck]): Custom checks listed here will be done in
-            addition to checks defined in the Data Package standard.
+        extensions (list[Extensions]): Additional checks (called extensions)
```
martonvago:
```suggestion
        extensions (Extensions): Additional checks (called extensions)
```

src/check_datapackage/custom_check.py
```diff
@@ -46,6 +52,59 @@ class CustomCheck:
...
+    """Extensions to the standard checks.
+
+    This sub-class of `Config` defines extensions, i.e., additional checks
+    that supplement those specified by the Data Package standard. It
+    contains sub-classes that store additional checks. This `Extensions` class
+    can be expanded to include more types of extensions.
+
+    Each extension class must implement its own `apply()` method that takes
+    the `datapackage.json` properties `dict` as input and outputs an `Issue`
+    list that contains the issues found by that extension.
```
martonvago: Do we want to formalise this requirement? As in, make the code express that all check classes must define `apply`?

martonvago (changes requested):
Nice, just some minor things

src/check_datapackage/custom_check.py
lwjohnst86: This will have to be in a separate PR. I'll make an issue for it.

martonvago (dismissed):
🎉

src/check_datapackage/custom_check.py
```diff
@@ -142,7 +199,9 @@ def apply_extensions(
...
         A list of `Issue`s.
     """
+    extensions_as_one = chain(extensions.required_checks, extensions.custom_checks)
     return _flat_map(
-        extensions,
-        lambda extension: extension.apply(properties),
+        extensions_as_one,
+        # TODO: fix type ignore when we use Protocols/ABCs
+        lambda extension: extension.apply(properties),  # type: ignore[attr-defined]
     )
```
martonvago:
```suggestion
    all_extensions: list[RequiredCheck | CustomCheck] = (
        extensions.required_checks + extensions.custom_checks
    )
    return _flat_map(
        all_extensions,
        lambda extension: extension.apply(properties),
    )
```

If you declare the type explicitly, it will know that `apply` exists even without a Protocol or super class

martonvago (changes requested):
Nice, just one comment ☺️

src/check_datapackage/custom_check.py
lwjohnst86: Ah, for some reason I thought I tried that and it didn't register. But I must have used `chain()` not `+`. Seems to work now! :+1:

martonvago (approved):
🎉

PR #167, opened 2025-10-27
feat: :sparkles: implement `error=True` in `check()`

### LINKED ISSUE #61: Implement `error=True` setting in `check()`
When `error` is set to `True` in `check()`, we have to raise an error if any issues are found instead of returning the issues as a list.

lwjohnst86:
# Description

Now outputs an error. Eventually we will want to replace the error message with `explain()`, but this is a good starting point.

Closes #61

Needs an in-depth review.

## Checklist

- [x] Ran `just run-all`

### DIFF
--- a/src/check_datapackage/check.py
+++ b/src/check_datapackage/check.py
@@ -38,7 +38,7 @@ sys.excepthook = no_traceback_hook
 
 
 class DataPackageError(Exception):
-    """Convert Data Package issues to an error and hiding the traceback."""
+    """Convert Data Package issues to an error and hide the traceback."""
 
     def __init__(
         self,
--- a/src/check_datapackage/check.py
+++ b/src/check_datapackage/check.py
@@ -50,15 +50,11 @@ class DataPackageError(Exception):
             issues,
             lambda issue: f"- Property `{issue.jsonpath}`: {issue.message}\n",
         )
-        self.message = (
+        message: str = (
             "There were some issues found in your `datapackage.json`:\n\n"
             + "\n".join(errors)
         )
-        super().__init__(self.message)
-
-    def __str__(self) -> str:
-        """Message to show when outputting the exception."""
-        return f"{self.message}"
+        super().__init__(message)
 
 
 def check(
--- a/src/check_datapackage/check.py
+++ b/src/check_datapackage/check.py
@@ -1,4 +1,5 @@
 import re
+import sys
 from dataclasses import dataclass
 from functools import reduce
 from typing import Any, Iterator, Optional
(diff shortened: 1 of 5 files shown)


src/check_datapackage/check.py
```diff
@@ -47,6 +48,19 @@ class for more details, especially about the default values.
...
     issues = exclude(issues, config.exclusions, properties)
 
+    if error and issues:
+        # TODO: Switch to using `explain()` once implemented
+        errors: list[str] = _map(
+            issues,
+            lambda issue: f"- Property `{issue.jsonpath}`: {issue.message}\n",
+        )
+        # Should hide the traceback for the user
+        sys.tracebacklimit = 0
```
martonvago: Is this a global setting? If so, do we not want to hide the traceback only here (using `raise ... from None`)?

martonvago (commented):
Nice, just some questions

src/check_datapackage/check.py
martonvago: Hmm, but doesn't this modify the interpreter instance? So anything that runs on the same interpreter (after this line) will be affected by this: errors inside `check-datapackage` but also in any external project that uses `check-datapackage` as a dependency will have no traceback.

At least if I import `check_datapackage` in another project as a dependency, trigger this line, and then raise an error in my own project code, then the traceback will be removed from that error as well.

lwjohnst86: Ugh, I dislike Python's "modify global settings in a function scope"... I'll try another solution, I had seen another option.

src/check_datapackage/check.py
```diff
@@ -18,6 +20,28 @@
 from check_datapackage.read_json import read_json
 
 
+def _no_traceback_hook(
+    exc_type: type[BaseException],
+    exc_value: BaseException,
+    exc_traceback: TracebackType | None,
+) -> None:
+    if IssuesAsErrors in exc_type.__bases__:
```
martonvago: I don't think this evaluates to true when `exc_type` is `IssuesAsErrors` (the traceback is not cut for me). Could just check that directly?

src/check_datapackage/check.py
```diff
@@ -47,6 +71,17 @@ class for more details, especially about the default values.
...
+    if error and issues:
+        # TODO: Switch to using `explain()` once implemented
+        errors: list[str] = _map(
+            issues,
+            lambda issue: f"- Property `{issue.jsonpath}`: {issue.message}\n",
+        )
+        raise IssuesAsErrors(
+            "There were some issues found in your `datapackage.json`:\n\n"
+            + "\n".join(errors)
+        )
```
martonvago: You could move the logic for constructing the error message inside `IssuesAsErrors`

martonvago (changes requested):

src/check_datapackage/check.py
```diff
@@ -20,26 +20,45 @@
...
 
-def _no_traceback_hook(
+def no_traceback_hook(
     exc_type: type[BaseException],
     exc_value: BaseException,
     exc_traceback: TracebackType | None,
 ) -> None:
-    if IssuesAsErrors in exc_type.__bases__:
+    """Exception hook to hide tracebacks for DataPackageError."""
+    if issubclass(exc_type, DataPackageError):
```
martonvago: Nice!

martonvago (changes requested):
Super nice!! 
Just a minor simplification if you want, otherwise 👍 👍

src/check_datapackage/check.py
```diff
@@ -20,26 +20,45 @@
...
+        )
+        self.message = (
+            "There were some issues found in your `datapackage.json`:\n\n"
+            + "\n".join(errors)
+        )
+        super().__init__(self.message)
+
+    def __str__(self) -> str:
+        """Message to show when outputting the exception."""
+        return f"{self.message}"
```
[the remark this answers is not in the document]
lwjohnst86: Good question. I just found it via googling.

martonvago (approved):
🎉 ⭐

PR #171, opened 2025-10-28
docs: :pencil2: use `properties`, not `descriptor` in Interface docs

lwjohnst86:
# Description

Needs a quick review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`

### DIFF
--- a/docs/design/interface.qmd
+++ b/docs/design/interface.qmd
@@ -32,8 +32,7 @@ This section describes the inputs accepted by `check-datapackage`.
     properties of a Data Package or Data Resource. Allowed properties
     are defined in the [Data Package
     standard](https://datapackage.org/). A common use case is loading
-    the properties from the `datapackage.json` file (which is called
-    a descriptor).
+    the properties from the `datapackage.json` file.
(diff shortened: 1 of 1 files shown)


docs/design/interface.qmd
```diff
@@ -28,11 +28,12 @@ components, along with their descriptions.
...
 This section describes the inputs accepted by `check-datapackage`.
 
--   **Descriptor:** This is a Python dictionary containing the
+-   **Properties**: This is a Python dictionary containing the
     properties of a Data Package or Data Resource. Allowed properties
     are defined in the [Data Package
     standard](https://datapackage.org/). A common use case is loading
-    the descriptor from the `datapackage.json` file.
+    the properties from the `datapackage.json` file (which is called
+    a descriptor).
```
signekb:
```suggestion
    the properties from the `datapackage.json` file.
```
Don't think we need that here if we're removing mentions of "descriptor" (almost?) everywhere.

signekb (changes requested):
👌 A few suggestions:

signekb (approved):

PR #175, opened 2025-10-29
feat: :sparkles: handle grouped errors under resource fields

martonvago:
# Description

This PR adds a function for handling grouped errors under `$.resources[x].schema.fields[x]`.
Here, the problem comes from the fact that each field type has its own sub-JSON-schema, and each one of these sub-schemas flags issues when the type of a field is not its own type. So, if a field has `type="number"` and there is something wrong with the field, then the sub-schemas for `year`, `string`, etc. will also flag issues. The goal is to flag issues only for `number`.

Part of #15 

Needs an in-depth review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`


src/check_datapackage/check.py
```diff
@@ -219,13 +231,52 @@ def _handle_S_resources_x_path(
...
+    # The field's type is known; keep only errors for this field type
+    schema_index = FIELD_TYPES.index(field_type)
+    errors_for_other_types = _filter(
+        errors_in_group,
+        lambda error: f"fields/items/oneOf/{schema_index}/" not in error.schema_path,
+    )
+    return _filter(
+        schema_errors,
+        lambda error: error not in errors_for_other_types,
+    )
```
martonvago: I'm not the biggest fan of passing `schema_errors` as an arg and then filtering it this way and that.
Maybe I'll eventually refactor this so that each handler function returns a list of errors to add and a list of errors to remove...

src/check_datapackage/check.py
```diff
@@ -116,6 +120,7 @@ class SchemaError:
     type: str
     schema_path: str
     jsonpath: str
+    instance: Any
```
lwjohnst86: Needs a docstring for it :relaxed:

lwjohnst86 (changes requested):
Nice! Only some small comments :relaxed:

lwjohnst86 (changes requested):
Very small change to the error message

src/check_datapackage/check.py
```diff
@@ -212,11 +220,54 @@ def _handle_S_resources_x_path(
...
+    errors_in_group = _get_errors_in_group(schema_errors, parent_error)
+    edits.remove.append(parent_error)
+
+    field_type: str = parent_error.instance.get("type", "string")
+
+    # The field's type is unknown
+    if field_type not in FIELD_TYPES:
+        unknown_field_error = SchemaError(
+            message=(
+                "Unknown Data Package field type. Please use one of"
```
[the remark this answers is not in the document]
martonvago: I think it is correct, it's just that
- there is no `field` property, we only have `fields`. This reads a bit like there was only one field
- the field property not having the correct type to me means that we have e.g. `"field": 123` instead of `"field": {...}`. But here we are talking about `"fields": [{"type":"...", ...}, ...]`

Maybe: "The type property of this resource schema field is incorrect. It should be one of..."? Because from the error message / issue it is clear which field this refers to.

src/check_datapackage/check.py
```diff
@@ -254,11 +262,54 @@ def _handle_S_resources_x_path(
...
+    edits.remove.append(parent_error)
+
+    field_type: str = parent_error.instance.get("type", "string")
+
+    # The field's type is unknown
+    if field_type not in FIELD_TYPES:
+        unknown_field_error = SchemaError(
+            message=(
+                "The type property in this resource schema field is incorrect. "
+                f"The value can only be one of these types: {', '.join(FIELD_TYPES)}."
```
lwjohnst86: Is that a missing `}`? Should it be:

```suggestion
                f"The value can only be one of these types: {', '.join(FIELD_TYPES)', '}.'"
```

lwjohnst86 (changes requested):

lwjohnst86 (approved):
:+1:

PR #183, opened 2025-11-04
feat: ✨ handle grouped errors under the enum constraint

martonvago:
# Description

This PR handles grouped errors at `$.resources[x].schema.fields[x].constraints.enum`. It's the worst one yet!

This schema node is deeply nested, so handling errors here is a little more complex.

The situation is the following: for each field, you can list the allowed values for that field under `"constraints": {"enum": [...]}`. Depending on the type of the field, the type of the values in `enum` can be different. Sometimes it's straightforward (e.g. for string fields, enum values have to be strings), but sometimes the schema for `enum` is more permissive (e.g. for number fields, enum values can be either numbers or strings). However, you cannot mix types.

If `enum` has mixed types or if some of the values are the wrong type, all sub-schemas under `enum` flag errors. These are usually contradictory, e.g., for a number field `[1, "inf"]` will alert both because 1 is not a string and because "inf" is not a number...

The logic in the function:
- if the error is not even for the current field type -> remove all errors
- if there are any errors on the `enum` property itself (not an array, duplicate items, etc.) -> keep only these
- the remaining errors are all about the individual values in the enum array being the wrong type
  -  if the values are not all the same type -> flag only this
  - if they are all the same type -> flag that this is the wrong type

Part of #15 

Needs an in-depth review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`


src/check_datapackage/check.py
```diff
@@ -304,12 +306,72 @@ def _handle_S_resources_x_schema_fields_x(
...
+        if field_type not in FIELD_TYPES:
+            return False
+        schema_index = FIELD_TYPES.index(field_type)
+        return f"fields/items/oneOf/{schema_index}/" in parent_error.schema_path
+
+    edits = SchemaErrorEdits(remove=[parent_error])
+    errors_in_group = _get_errors_in_group(schema_errors, parent_error)
+
+    # Remove errors for other field types
+    if not _error_is_for_field_type():
```
```suggestion
    if _is_not_error_for_field_type(parent_error):
```

src/check_datapackage/check.py
```diff
@@ -304,12 +306,72 @@ def _handle_S_resources_x_schema_fields_x(
     return edits
 
 
+def _handle_S_resources_x_schema_fields_x_constraints_enum(
+    parent_error: SchemaError,
+    schema_errors: list[SchemaError],
+) -> SchemaErrorEdits:
+    """Only flag errors for the relevant field type and simplify errors."""
+
+    def _error_is_for_field_type() -> bool:
```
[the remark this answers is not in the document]
lwjohnst86: Also this function should really be moved out on its own. Nested functions make a function a bit cluttered.

src/check_datapackage/check.py
```diff
@@ -304,12 +306,72 @@ def _handle_S_resources_x_schema_fields_x(
...
+        if not parent_error.parent:
+            return False
```
```suggestion
            return True
```

src/check_datapackage/check.py
```diff
@@ -304,12 +306,72 @@ def _handle_S_resources_x_schema_fields_x(
...
+        field_type: str = parent_error.parent.instance.get("type", "string")
+        if field_type not in FIELD_TYPES:
+            return False
```
```suggestion
            return True
```

src/check_datapackage/check.py
```diff
@@ -304,12 +306,72 @@ def _handle_S_resources_x_schema_fields_x(
...
+        schema_index = FIELD_TYPES.index(field_type)
+        return f"fields/items/oneOf/{schema_index}/" in parent_error.schema_path
+
+    edits = SchemaErrorEdits(remove=[parent_error])
+    errors_in_group = _get_errors_in_group(schema_errors, parent_error)
+
+    # Remove errors for other field types
+    if not _error_is_for_field_type():
+        edits.remove.extend(errors_in_group)
+        return edits
```
lwjohnst86: Hmm, there are a too many returns in this function, which suggests maybe it's doing too much. Could you redesign in a way to not have so many early exits?

lwjohnst86 (changes requested):
Neat! Just some fairly small comments :relaxed:

src/check_datapackage/check.py
```diff
@@ -304,12 +306,72 @@ def _handle_S_resources_x_schema_fields_x(
...
+
+    # Keep only top-level enum errors, if any
+    if enum_errors:
+        return edits
+
+    # Replace value errors with a simpler error
+    message = "All enum values must be the same type."
+    same_type = len(set(_map(parent_error.instance, lambda value: type(value)))) == 1
+    if same_type:
+        allowed_types = set(_map(value_errors, lambda error: str(error.schema_value)))
```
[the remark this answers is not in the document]
martonvago: Because there can be more errors than allowed types. 
E.g., if the allowed types are array and object and the enum array is `[1, 2]`, then both enum items will cause errors for both allowed types. So 4 errors but only 2 allowed types.

src/check_datapackage/check.py
```diff
@@ -156,6 +156,7 @@ class SchemaError:
             Path components are separated by '/'.
         jsonpath (str): The JSON path to the field that violates the check.
         instance (Any): The part of the object that failed the check.
+        schema_value (Optional[Any]): The part of the schema violated by this error.
```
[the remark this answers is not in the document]
lwjohnst86: How is this different from `jsonpath`? The explanation/description reads very similarly to me.

src/check_datapackage/check.py
```diff
@@ -304,6 +306,64 @@ def _handle_S_resources_x_schema_fields_x(
...
+def _get_unified_enum_values_error(
+    parent_error: SchemaError,
+    value_errors: list[SchemaError],
+) -> SchemaError:
+    message = "All enum values must be the same type."
+    same_type = len(set(_map(parent_error.instance, lambda value: type(value)))) == 1
+    if same_type:
+        allowed_types = set(_map(value_errors, lambda error: str(error.schema_value)))
+        message = (
+            "Incorrect enum value type. Enum values should be "
```
```suggestion
            "The enum value type is not correct. Enum values should be "
```

Small change to make it a bit less formal sounding (to my ears anyway)

src/check_datapackage/check.py
```diff
@@ -304,6 +306,64 @@ def _handle_S_resources_x_schema_fields_x(
...
+def _handle_S_resources_x_schema_fields_x_constraints_enum(
+    parent_error: SchemaError,
+    schema_errors: list[SchemaError],
+) -> SchemaErrorEdits:
+    """Only flag errors for the relevant field type and simplify errors."""
+    edits = SchemaErrorEdits(remove=[parent_error])
+    errors_in_group = _get_errors_in_group(schema_errors, parent_error)
+
+    # Remove errors for other field types
+    if _error_not_for_field_type(parent_error):
```
```suggestion
    if _not_field_type_error(parent_error):
```

Sounds a bit better like this.

src/check_datapackage/check.py
```diff
@@ -304,6 +306,64 @@ def _handle_S_resources_x_schema_fields_x(
...
+    return SchemaError(
+        message=message,
+        type="type",
+        schema_path=value_errors[0].schema_path,
+        jsonpath=_strip_index(value_errors[0].jsonpath),
+        instance=value_errors[0].instance,
+    )
+
+
+def _error_not_for_field_type(parent_error: SchemaError) -> bool:
```
```suggestion
def _not_field_type_error(parent_error: SchemaError) -> bool:
```

lwjohnst86 (changes requested):
Nice! Some fairly small comments :+1:

tests/test_check.py
```diff
@@ -389,28 +484,6 @@ def test_fail_foreign_keys_with_missing_fields(ref_fields):
...
-    properties = example_package_properties()
-    properties["resources"][0]["licenses"] = [{}]
-
-    issues = check(properties)
-
-    assert len(issues) == 1
-    assert issues[0].type == "required"
-    assert issues[0].jsonpath == "$.resources[0].licenses[0]"
-
-
```
martonvago: Moved these up to tidy the order of tests

lwjohnst86 (changes requested):
Very minor suggestion

lwjohnst86 (approved):

PR #184, opened 2025-11-05
feat: ✨ use Pydantic in config classes

### LINKED ISSUE #173: Use Pydantic to verify config settings?
> From the point of view of us, the user, whether the check happens with `__post_init__()` or in `apply()`, it doesn't really matter, since we run the script, not the lines. We'll get an error message anyway when we use it in the script. Where exactly it happens doesn't matter so long as the error message is informative enough. So if it doesn't matter from the point of view of runtime, I don't see how the added complexity is necessary. Especially since `__post_init__()` method signature doesn't tell me the reader what it's doing. The fact it is doing a check isn't obvious from the signature, so to me, this is more of a hack of the way classes work in Python than an actual action of checking the config input. 
> 
> But, I do see your point that we should have some verification going on with the way that we (eventually) read the config settings. For example, restricting `version` to only correct values like "v2" etc. I was looking to config file parsing and many people mention using Pydantic to confirm that the config is correct to how we want it.
> 
> Given this is a bigger issue than adding the `RequiredCheck` functionality, can you remove these checks from `CustomCheck` and `RequiredCheck` from this PR? So that the PR stays atomic (right now it is drifting a lot from being atomic). And then I'll open an issue (with this comment thread) for us to figure out how to verify the config settings. Then we can decide if Pydantic makes more sense or if there is another tool. This also might mean we will have to switch from using dataclasses to using Pydantic's models.

_Originally posted by @lwjohnst86 in https://github.com/seedcase-project/check-datapackage/pull/122#discussion_r2472110594_

martonvago:
# Description

This PR makes the config classes into Pydantic models with the appropriate checks.

Pydantic allows lots of configuration and in many different ways, so this is just a starting point.

Closes #173 

Needs an in-depth review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`


src/check_datapackage/config.py
```diff
@@ -47,7 +47,7 @@ class Config:
         ```
     """
 
-    exclusions: list[Exclusion] = field(default_factory=list[Exclusion])
+    exclusions: list[Exclusion] = []
```
martonvago: Pydantic should [take care](https://docs.pydantic.dev/latest/concepts/fields/#mutable-default-values) of this

src/check_datapackage/extensions.py
```diff
@@ -49,11 +51,21 @@ class CustomCheck:
...
+    @field_validator("type", mode="after")
+    @classmethod
+    def _check_not_required(cls, value: str) -> str:
+        if value == "required":
+            raise ValueError(
+                "Cannot define `CustomCheck` with `type='required'`."
+                " Use `RequiredCheck` to mark fields as required instead."
+            )
+        return value
+
```
martonvago: This is a decorator that will run after built-in Pydantic checks are run. So `value` is guaranteed to be a string at this point.

src/check_datapackage/extensions.py
```diff
@@ -99,8 +110,24 @@ class RequiredCheck:
...
+        if not field_name_match:
+            raise ValueError(
+                f"Cannot define `RequiredCheck` for JSON path `{self.jsonpath}`."
+                " A `RequiredCheck` must target a concrete object field (e.g.,"
+                " `$.title`) or set of fields (e.g., `$.resources[*].title`)."
+                " Ambiguous paths (e.g., `$..title`) or paths pointing to array items"
+                " (e.g., `$.resources[0]`) are not allowed."
+            )
+
+        object.__setattr__(self, "_field_name", field_name_match.group(1))
```
martonvago: We are back to this question: should `_field_name` be saved or recomputed?

If recomputed, this model validator will look nicer, but then we can't delete the now deleted lines in `apply`. 🤔 

For what it's worth, Pydantic seems explicitly to allow modifying the object in its validators.

src/check_datapackage/extensions.py
```diff
@@ -49,11 +51,21 @@ class CustomCheck:
...
     message: str
     check: Callable[[Any], bool]
     type: str = "custom"
 
+    @field_validator("type", mode="after")
+    @classmethod
+    def _check_not_required(cls, value: str) -> str:
+        if value == "required":
+            raise ValueError(
+                "Cannot define `CustomCheck` with `type='required'`."
```
lwjohnst86:
```suggestion
                "Cannot use `CustomCheck` with `type='required'`."
```

src/check_datapackage/extensions.py
```diff
@@ -99,8 +110,24 @@ class RequiredCheck:
         ```
     """
 
-    jsonpath: str
+    jsonpath: JsonPath
     message: str
+    _field_name: str = PrivateAttr()
+
+    @model_validator(mode="after")
```
lwjohnst86: Why is `classmethod` used above but not here?

src/check_datapackage/extensions.py
```diff
@@ -99,8 +110,24 @@ class RequiredCheck:
...
+    def _check_field_name_in_jsonpath(self) -> Self:
+        field_name_match = re.search(r"(?<!\.)(\.\w+)$", self.jsonpath)
+        if not field_name_match:
+            raise ValueError(
+                f"Cannot define `RequiredCheck` for JSON path `{self.jsonpath}`."
```
lwjohnst86:
```suggestion
                f"Cannot use `RequiredCheck` for this JSON path `{self.jsonpath}`."
```

src/check_datapackage/extensions.py
```diff
@@ -99,8 +110,24 @@ class RequiredCheck:
...
+                " A `RequiredCheck` must target a concrete object field (e.g.,"
+                " `$.title`) or set of fields (e.g., `$.resources[*].title`)."
+                " Ambiguous paths (e.g., `$..title`) or paths pointing to array items"
```
lwjohnst86:
```suggestion
                " Ambiguous or indirect JSON paths (e.g., `$..title`) or JSON paths pointing to array items"
```

src/check_datapackage/extensions.py
lwjohnst86: I don't love it, but with the use of `PrivateAttr()` and Pydantic in general, this is ok.

lwjohnst86 (changes requested):
Very nice :tada: just some mostly small comments/suggestions.

src/check_datapackage/extensions.py
martonvago: Because this is a regular instance method (i.e. it is tied to the concrete instance and knows about its fields). Model validators are instance methods and field validators are class methods in the Pydantic docs.

src/check_datapackage/extensions.py
```diff
@@ -99,8 +110,24 @@ class RequiredCheck:
...
     message: str
+    _field_name: str = PrivateAttr()
+
+    @model_validator(mode="after")
+    def _check_field_name_in_jsonpath(self) -> Self:
+        field_name_match = re.search(r"(?<!\.)(\.\w+)$", self.jsonpath)
+        if not field_name_match:
+            raise ValueError(
+                f"Cannot define `RequiredCheck` for JSON path `{self.jsonpath}`."
+                " A `RequiredCheck` must target a concrete object field (e.g.,"
```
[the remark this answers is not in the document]
lwjohnst86: I think "specific" sounds the best. What about "must use a JSON path that targets a specific and real property"?

lwjohnst86 (changes requested):
Other than my last comment, this looks good!

lwjohnst86 (approved):

PR #185, opened 2025-11-07
feat: :sparkles: handle grouped errors for primary key

martonvago:
# Description

This PR handles grouped errors under `$.resources[x].schema.primaryKey`.
This is a simpler situation as primary keys have to be strings or arrays of strings.

Part of #15

Needs an in-depth review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`


src/check_datapackage/check.py
```diff
@@ -304,12 +304,47 @@ def _handle_S_resources_x_schema_fields_x(
     return edits
 
 
+def _handle_S_resources_x_schema_primary_key(
+    parent_error: SchemaError,
+    schema_errors: list[SchemaError],
+) -> SchemaErrorEdits:
+    """Only flag errors for the relevant type and simplify errors."""
+    PRIMARY_KEY_TYPES: tuple[type[Any], ...] = (list, str)
```
martonvago: I left this in this function because nothing else needs to know about it.

I expressed it rather generically, technically, all we need to know is that the schema index of the list/array-based schema is 0.

lwjohnst86 (approved):
Amazing :tada:

lwjohnst86 (dismissed):

lwjohnst86:
@martonvago just need to resolve conflicts :mechanical_arm:

lwjohnst86 (approved):

PR #192, opened 2025-11-13
feat: :sparkles: expose `DataPackageError`

martonvago:
# Description

This PR exposes `DataPackageError`. We raise this error so consumers should be able to catch it.

Needs a quick review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`


lwjohnst86 (approved):
:+1:

PR #195, opened 2025-11-17
docs: :memo: revise guide docs to use `example_*` (plus minor edits)

### LINKED ISSUE #164: Use `example_*` functions in guide docs
Rather than a custom dict.

lwjohnst86:
# Description

Some minor edits after converting the guide docs to use `example_package_properties()`

Closes #164

Needs an in-depth review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`

### DIFF
--- a/docs/guide/config.qmd
+++ b/docs/guide/config.qmd
@@ -4,9 +4,10 @@ jupyter: python3
 order: 3
 ---
 
-You can pass a `Config` object to `check()` to customise the checks done
-on your Data Package's properties. The following configuration options
-are available:
+You can pass a [`Config`](/docs/reference/Config.qmd) object to
+[`check()`](/docs/reference/check.qmd) to customise the checks done on
+your Data Package's properties. The following configuration options are
+available:
 
 -   `version`: The version of Data Package standard to check against.
     Defaults to `v2`.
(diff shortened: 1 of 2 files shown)


docs/guide/config.qmd
```diff
@@ -39,73 +40,82 @@ the `required` check by defining an `Exclusion` object with this `type`:
...
 ```{python}
 exclusion_path = cdp.Exclusion(jsonpath="$.resources[*].path")
+exclusion_path
 ```
 
-The `type` and `jsonpath` arguments can also be combined:
+The `type` and `jsonpath` arguments can also be combined, so we can
+ignore [`Issue`](/docs/reference/Issue.qmd) from a specific type on a
+specific field. For example, to exclude checks that the `created` field
+be in a date format (`type="format"`), we can use:
```
martonvago:
```suggestion
specific field. For example, to exclude checks that verify the `created` field
is in a date format (`type="format"`), we can use:
```

docs/guide/config.qmd
```diff
@@ -39,73 +40,82 @@ the `required` check by defining an `Exclusion` object with this `type`:
...
-the package and resource properties, and the resource `path` doesn't
-point to a data file. However, as we have defined exclusions for all of
-these, the function will flag no issues.
+When we run `check()` on these properties, we get the four issues:
+
+```{python}
+cdp.check(properties=package_properties)
+```
+
+Now let's exclude these Issues so that `check()` finds no issues by
```
martonvago:
```suggestion
Now let's exclude these `Issue`s so that `check()` finds no issues by
```

martonvago (changes requested):
Nice, just very minor things ⭐

docs/guide/config.qmd
```diff
@@ -39,73 +40,82 @@ the `required` check by defining an `Exclusion` object with this `type`:
...
+package_properties["resources"][0]["path"] = "\\not/a/path"
+pprint.pp(package_properties)
 ```
 
-In the example above, we would expect four `Issue` items: the package
-`name` is a number, the required `description` field is missing in both
-the package and resource properties, and the resource `path` doesn't
-point to a data file. However, as we have defined exclusions for all of
-these, the function will flag no issues.
+When we run `check()` on these properties, we get the four issues:
```
signekb:
```suggestion
When we run `check()` on these properties, we get the three expected issues:
```

signekb (dismissed):
👌 👌  A couple of suggestions:

martonvago (approved):

PR #199, opened 2025-11-17
docs: :lipstick: prettify the guide list in the landing page

lwjohnst86:
# Description

Makes it a list of items, rather than a row of them. Adds description to fill in the space a bit.

Needs a quick review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`

### DIFF
--- a/docs/guide/config.qmd
+++ b/docs/guide/config.qmd
@@ -1,5 +1,6 @@
 ---
 title: "Configuring the checks"
+description: "Customise the types of checks that are done against your `datapackage.json` by using the `Config` class, such as excluding certain checks or adding your own."
 jupyter: python3
 order: 3
 ---
--- a/docs/guide/config.qmd
+++ b/docs/guide/config.qmd
@@ -1,6 +1,6 @@
 ---
 title: "Configuring the checks"
-description: "Customise the types of checks that are done against your `datapackage.json` by using the `Config` class, such as excluding certain checks or adding your own."
+description: "A simple introduction to customising the types of checks that are done against your `datapackage.json` by using the `Config` class, such as excluding certain checks or adding your own."
 jupyter: python3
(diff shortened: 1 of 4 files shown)


docs/guide/index.qmd
```diff
@@ -4,7 +4,10 @@ description: "A guide on using `check-datapackage` to check the compliance of yo
 listing:
     contents: .
     type: grid
+    grid-columns: 1
+    grid-item-border: false
```
joelostblom: Personally, I prefer including the borders. Without them, each items seems less like a separate clickable heading to me and instead blends together with the other headings on the page:

Without border (current):
Image

With border:
Image

docs/guide/config.qmd
```diff
@@ -1,5 +1,6 @@
 ---
 title: "Configuring the checks"
+description: "Customise the types of checks that are done against your `datapackage.json` by using the `Config` class, such as excluding certain checks or adding your own."
```
joelostblom: It would look aesthetically pleasing if each of these descriptions were written in the same format. The other two you edited started in the same manner: "A brief walkthrough..." and "A short description", whereas this one and the installation one follows a different format starting with a verb and including more of the content in the description. I don't have a strong preference for which way these are written, but think it would look nice if they were more similar.

joelostblom (changes requested):
I like this format better than the multi-column grid! I was actually going to suggest removing it altogether because I found it redundant with the ToC on the left, but the change here with the additional description adds some value and makes it easier to see that this is the same as the ToC because the order and formatting is the same (ie one vertical column)

signekb (commented):
Nice! 💄 A comment:

joelostblom (approved):
Looks good!

PR #201, opened 2025-11-19
feat: :sparkles: support union operator in `RequiredCheck`

### LINKED ISSUE #193: [discussion]: Support union JSON path operator in `RequiredCheck` as well?
In Sprout, we often want the same check for fields at many different JSON paths.

In JSON path notation, you can express the union of paths in two ways:
1. `$.name | $.title`
2. `$['name', 'title']`

Either of these will return 2 matches: name and title.

This means we can write one `Exclusion` like
```python
Exclusion(jsonpath="$.name | $.title", type="required")
```
to exclude required issues at both paths.
The same is true for `CustomCheck`.

We disabled this for `RequiredCheck` in the first pass for simplicity (also, I hadn't looked into it yet fully), but we could allow it. Should we?

martonvago:
# Description

This PR adds support for the union JSON path operator for `RequiredCheck`s.

To be able to do this more intelligently than a monster regex, I looked into how `jsonpath` represents a JSON path internally and used their terms/classification. This also allowed me to be more precise about what JSON paths we are not allowing for `RequiredCheck`s.

Closes #193 

Needs an in-depth review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`

### DIFF
--- a/src/check_datapackage/extensions.py
+++ b/src/check_datapackage/extensions.py
@@ -1,5 +1,6 @@
 from collections.abc import Callable
 from dataclasses import dataclass
+from operator import itemgetter
 from typing import Any, Self, cast
 
 from jsonpath import JSONPath, compile
@@ -120,7 +121,7 @@ def _jsonpath_to_targets(jsonpath: JSONPath) -> list[TargetJsonPath]:
             f"Cannot use the JSON path `{full_path}` in `RequiredCheck`"
             " because it ends in the recursive descent (`..`) operator."
         )
-    
+
     # Things like field names, array indices, and/or wildcards.
     selectors = last_segment.selectors
     if _filter(selectors, lambda selector: not isinstance(selector, NameSelector)):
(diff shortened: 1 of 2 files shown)


src/check_datapackage/extensions.py
```diff
@@ -112,22 +158,20 @@ class RequiredCheck(BaseModel, frozen=True):
...
-                " not allowed."
-            )
-
-        object.__setattr__(self, "_field_name", field_name_match.group(1))
+        jsonpath = compile(self.jsonpath)
+        if isinstance(jsonpath, JSONPath):
+            paths = [jsonpath]
+        else:
+            first_path = cast(JSONPath, jsonpath.path)
+            paths = [first_path] + _map(jsonpath.paths, lambda path: path[1])
```
lwjohnst86:
```suggestion
            paths = [first_path] + _map(jsonpath.paths, itemgetter(0))
```

Rather than use lambda. Needs to import `from operator import itemgetter`

src/check_datapackage/extensions.py
```diff
@@ -112,22 +158,20 @@ class RequiredCheck(BaseModel, frozen=True):
...
+
+        object.__setattr__(
+            self, "_targets", _flat_map(paths, lambda path: _jsonpath_to_targets(path))
```
lwjohnst86:
```suggestion
            self, "_targets", _flat_map(paths, _jsonpath_to_targets)
```

No need for lambda

src/check_datapackage/extensions.py
```diff
@@ -91,6 +94,49 @@ def apply(self, properties: dict[str, Any]) -> list[Issue]:
...
+        return []
+
+    full_path = jsonpath.segments[0].token.path
+    last_segment = jsonpath.segments[-1]
+    if isinstance(last_segment, JSONPathRecursiveDescentSegment):
+        raise ValueError(
+            f"Cannot use `RequiredCheck` for the JSON path `{full_path}`"
+            " because it ends in the recursive descent (`..`) operator."
+        )
+
```
lwjohnst86:
```suggestion
    
    # Things like field names, array indices, and/or wildcards.
```

lwjohnst86 (changes requested):
Awesome possum :rabbit: some small edits

lwjohnst86 (approved):

PR #208, opened 2025-11-21
feat: :sparkles: `explain()` errors

### LINKED ISSUE #11: Brainstorming for design of messages
Put ideas here. Obviously we are constrained by what is possible in Python, but this is to discuss some ideas for improving the messages of the issues detected.

### LINKED ISSUE #14: Improve `CheckError` error messages
Following Luke's suggestions, we should improve the error messages generated by `jsonschema` and exposed by the `checks` package.

So, in `checks`, we should do the following modifications to all errors before returning them:
- Type error message to: `f"The '{field}' field needs to be a 'string'."`
- Required error message to: `f"The '{field}' field needs to have a value in it, it is required."`

Some options:
1. In `validation_errors_to_check_errors`, have a custom function generate the error message based on the `ValidationError`.
2. Have `CheckError` itself generate the error message based on the `validator` and the name of the field when the error in constructed.

Make sure that `CheckError`s raised in Sprout use these same error messages.

### LINKED ISSUE #58: Create `explain()` function

joelostblom:
Creates a more human readable format to explain error messages

Closes #11, closes #58, closes #14

Needs an in-depth review.

## Checklist

- [x] Ran `just run-all`

### DIFF
--- a/src/check_datapackage/check.py
+++ b/src/check_datapackage/check.py
@@ -602,7 +602,11 @@ def _create_issue(error: SchemaError) -> Issue:
         message=error.message,
         jsonpath=error.jsonpath,
         type=error.type,
-        instance=error.instance,
+        instance=(
+            None
+            if error.type in ("required", "minItems", "oneOf", "enum", "uniqueItems")
+            else error.instance,
+        ),
     )
 
 
--- a/src/check_datapackage/check.py
+++ b/src/check_datapackage/check.py
@@ -55,7 +55,7 @@ class DataPackageError(Exception):
             explain,
         )
         message: str = (
-            "There were some issues found in your `datapackage.json`:\n\n"
+            f"There were {len(errors)} issues found in your `datapackage.json`:\n\n"
             + "\n".join(errors)
         )
         super().__init__(message)
--- a/src/check_datapackage/check.py
+++ b/src/check_datapackage/check.py
@@ -78,8 +78,10 @@ def explain(issues: list[Issue]) -> str:
         issues,
         _create_explanation,
     )
+    num_issues = len(issue_explanations)
+    singular_or_plural = " was" if num_issues == 1 else "s were"
     return (
-        f"{len(issue_explanations)} issues were found in your `datapackage.json`:\n\n"
+        f"{num_issues} issue{singular_or_plural} found in your `datapackage.json`:\n\n"
         + "\n".join(issue_explanations)
     )
 
--- a/src/check_datapackage/check.py
+++ b/src/check_datapackage/check.py
@@ -3,7 +3,7 @@ import sys
 from dataclasses import dataclass, field
 from functools import reduce
 from types import TracebackType
-from typing import Any, Callable, Iterator, Optional
+from typing import Any, Callable, Hashable, Iterator, Optional
 
 from jsonschema import Draft7Validator, FormatChecker, ValidationError
 
@@ -602,11 +602,8 @@ def _create_issue(error: SchemaError) -> Issue:
         message=error.message,
         jsonpath=error.jsonpath,
         type=error.type,
-        instance=(
-            None
-            if error.type in ("required", "minItems", "oneOf", "enum", "uniqueItems")
-            else error.instance,
-        ),
+        # Since we want to deduplicate issues, only include instance if it's hashable
+        instance=error.instance if isinstance(error.instance, Hashable) else None,
     )
 
 
--- a/src/check_datapackage/check.py
+++ b/src/check_datapackage/check.py
@@ -52,13 +52,32 @@ class DataPackageError(Exception):
         # TODO: Switch to using `explain()` once implemented
         errors: list[str] = _map(
             issues,
-            lambda issue: f"- Property `{issue.jsonpath}`: {issue.message}\n",
+            explain,
         )
         message: str = (
             "There were some issues found in your `datapackage.json`:\n\n"
             + "\n".join(errors)
         )
         super().__init__(message)
+
+
+def explain(issue: Issue) -> str:
+    """Explain the issue in a human-readable format.
+
+    Args:
+        issue: The `Issue` object to explain.
+
+    Returns:
+        A human-readable explanation of the issue.
+    """
+    property_name = issue.jsonpath.split(".")[-1]
+    return (  # noqa: F401
+        f"At package.{issue.jsonpath[2:]}:\n"  # indexing to remove '$.'
+        "|\n"
+        f"| {property_name}: {issue.instance!r}\n"
+        f"| {' ' * len(property_name)}  {'^' * len(str(issue.instance))}\n"
+        f"{issue.message}\n"
+    )
 
 
 def check(
@@ -578,6 +597,7 @@ def _create_issue(error: SchemaError) -> Issue:
         message=error.message,
         jsonpath=error.jsonpath,
         type=error.type,
+        instance=error.instance,
     )
 
 
--- a/src/check_datapackage/check.py
+++ b/src/check_datapackage/check.py
@@ -66,7 +66,7 @@ def explain(issues: list[Issue]) -> str:
         _create_explanation,
     )
     error_message: str = (
-        f"There were {len(issue_explanations)} issues found in your `datapackage.json`:\n\n"
+        f"{len(issue_explanations)} issues were found in your `datapackage.json`:\n\n"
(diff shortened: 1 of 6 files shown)


src/check_datapackage/issue.py
```diff
@@ -31,3 +32,4 @@ class Issue:
     jsonpath: str
     type: str
     message: str
+    instance: Any
```
joelostblom: We could also enumerate the possible types here, but I'm not 100% on what they should be. From running

```sh
rg "\"type\": \"" ./src/check_datapackage/schemas/data-package-2-0.json |tr -d " " | tr -d ',' | sort | uniq
```

I see:

```
"type":"array"
"type":"boolean"
"type":"integer"
"type":"number"
"type":"object"
"type":"string"
```

It seems like `array` could be `List`, but what is `object`? Maybe `Dict`?

src/check_datapackage/check.py
```diff
@@ -52,15 +52,34 @@ def __init__(
...
+    Returns:
+        A human-readable explanation of the issue.
+    """
+    property_name = issue.jsonpath.split(".")[-1]
+    return (  # noqa: F401
+        f"At package.{issue.jsonpath[2:]}:\n"  # indexing to remove '$.'
+        "|\n"
+        f"| {property_name}: {issue.instance!r}\n"
+        f"| {' ' * len(property_name)}  {'^' * len(str(issue.instance))}\n"
+        f"{issue.message}\n"
```
lwjohnst86: I think we might need to have some custom messages, as some of the messages are not helpful (e.g. the regex ones).

But that can be another PR :relaxed:

lwjohnst86 (changes requested):
Super cool :tada: Some small comments so far :pen:

src/check_datapackage/check.py
```diff
@@ -48,17 +48,40 @@ def __init__(
...
+
+    Args:
+        issues: A list of `Issue` objects to explain.
+
+    Returns:
+        A human-readable explanation of the issue.
+    """
+    issue_messages: list[str] = _map(
+        issues,
+        _create_issue_message,
```
```suggestion
        _create_explain_message,
```

lwjohnst86 (changes requested):
Nice! Very minor comment about naming :relaxed:

src/check_datapackage/extensions.py
```diff
@@ -162,6 +163,7 @@ class RequiredCheck(BaseModel, frozen=True):
     jsonpath: JsonPath
     message: str
     _targets: list[TargetJsonPath] = PrivateAttr()
+    instance: Any = None
```
joelostblom: @martonvago @lwjohnst86 I set the default to `None` so that we don't have to pass `error.instance` around everywhere that we create an Issue (since it is only used for the `explain` message right now), but if you think that approach is better, I can change to that.

lwjohnst86 (changes requested):

lwjohnst86 (changes requested):

src/check_datapackage/check.py
```diff
@@ -579,6 +602,7 @@ def _create_issue(error: SchemaError) -> Issue:
         message=error.message,
         jsonpath=error.jsonpath,
         type=error.type,
+        instance=error.instance,
```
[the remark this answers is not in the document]
joelostblom: After some testing, it seems like any properties dictionary/json that misses the top level `resources` attribute triggers this error... :thinking:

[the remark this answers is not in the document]
lwjohnst86: Hmmm, maybe for those cases, don't include the instance and leave it `''` blank? :thinking:

src/check_datapackage/extensions.py
lwjohnst86: Hmm, thinking about this more, if `instance` is part of the `explain` message, we should actually require that it be present, no? Rather than default to `None`?

joelostblom: Ok, tomorrow I can look into the other places where it is used and see if it makes sense to always include a message there. For example for the required and [custom check classes](https://github.com/seedcase-project/check-datapackage/blob/9410ee4881207ae124d86aeacefe4a4e4154a550/src/check_datapackage/extensions.py#L92-L94) (maybe you know already if that makes sense?).

src/check_datapackage/check.py
```diff
@@ -48,17 +48,40 @@ def __init__(
...
+        + "\n".join(issue_explanations)
+    )
+    return error_message
+
+
+def _create_explanation(issue: Issue) -> str:
+    """Create an informative explanation of what went wrong in each issue."""
+    property_name = issue.jsonpath.split(".")[-1]
+    return (  # noqa: F401
+        f"At package.{issue.jsonpath[2:]}:\n"  # indexing to remove '$.'
```
```suggestion
        f"At `package.{issue.jsonpath[2:]}`:\n"  # indexing to remove '$.'
```

I think backticks or some other visual separator would be nice here.

src/check_datapackage/check.py
```diff
@@ -48,17 +48,40 @@ def __init__(
...
+        "|\n"
+        f"| {property_name}: {issue.instance!r}\n"
```
lwjohnst86: Should we wrap `property_name` in `""` to match JSON style or match Python style? :thinking:

src/check_datapackage/check.py
joelostblom: Personally, I think it can look a bit cluttered with raw markup being printed as output. Could we punt this to a PR that deals with the styles of these messages and considers using rich? E.g. just using the `print` from rich would give us some additional formatting:

image


Rich also does pretty printing of json objects and creates stylized tracebacks, such as this example:

image

src/check_datapackage/extensions.py
lwjohnst86: Needs a docstring for this attribute.

lwjohnst86 (changes requested):
Look super nice! Just some small comments about docstrings.

src/check_datapackage/check.py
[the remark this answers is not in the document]
martonvago: Yeah, exactly what I was thinking too! And yeah, I wasn't sure how you wanted to handle these bulky `instance`s in the explain output. If you don't want to show them at all, then weeding them out earlier is more practical. But, conceptually at least, removing `instance` from the hash and handling `instance`s we don't want to display as part of the display logic feels cleaner to me. It just feels a bit off to be exposing `instance` to the user and not have it reliably contain the instance. 🤔

src/check_datapackage/check.py
```diff
@@ -48,17 +48,55 @@ def __init__(
...
+    return error_message
+
+
+def _create_explanation(issue: Issue) -> str:
+    """Create an informative explanation of what went wrong in each issue."""
+    property_name = issue.jsonpath.split(".")[-1]
+    # Two extra carets are added for string instances to account for quotes
+    number_of_carets = len(str(issue.instance)) + 2 * isinstance(issue.instance, str)
+    return (  # noqa: F401
+        f"At package.{issue.jsonpath[2:]}:\n"  # indexing to remove '$.'
```
martonvago: Can also do `issue.jsonpath.removeprefix("$.")` and then maybe the comment is not necessary

src/check_datapackage/check.py
martonvago: Here, we rewrite `jsonpath` so that for `type="required"` errors it ends in the missing field's name. I think that (regardless of having solved the hash problem) this should be matched by rewriting `instance` for required errors in the same way. Otherwise, we will end up with output like:
```
# check({"resources": [{"title": "Title"}]}, error=True)

At package.resources[0].name:
|
| name: {'title': 'Title'}
|       ^^^^^^^^^^^^^^^^^^
'name' is a required property
```

It would be nice if this could happen here and not in `_create_issue`, but I don't know off the top of my head if that would be problematic for any logic that relies on `instance` on `SchemaError`s.

martonvago (dismissed):
Really nice ⭐!!

I added some comments, feel free to ignore what you don't care about ☺️

src/check_datapackage/check.py
joelostblom: Interesting, do you think this is something that can be solved in a separate PR? It's not immediately clear how to differentiate the two errors I get from running your example:

Image

The first one seems correct whereas the second one has the issue you highlighted.

joelostblom (commented):
Thanks @martonvago ! I went through and tried to address your comments.

lwjohnst86 (approved):
Very nice :tada:

PR #214, opened 2025-11-26
fix: :bug: add dict check when getting parent instance

martonvago:
# Description

This PR adds an extra check to make sure `parent_error.instance` is a `dict` when we want to access any properties on it. 

Needs a quick review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`

### DIFF
--- a/src/check_datapackage/check.py
+++ b/src/check_datapackage/check.py
@@ -274,11 +274,14 @@ def _handle_S_resources_x_schema_fields_x(
 
     E.g., if the field type is `string`, flag errors for the string-based schema only.
     """
-    edits = SchemaErrorEdits()
+    edits = SchemaErrorEdits(remove=[parent_error])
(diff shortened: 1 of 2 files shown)


src/check_datapackage/check.py
```diff
@@ -413,11 +416,14 @@ def _handle_S_resources_x_schema_foreign_keys(
     edits = SchemaErrorEdits(remove=[parent_error])
     errors_in_group = _get_errors_in_group(schema_errors, parent_error)
 
-    key = parent_error.instance.get("fields", None)
```
martonvago: I actually don't want the default to be `None` because then I conflate the cases when `fields` is not set and when it is set to `None`. The latter is flagged by `jsonschema`, so I want to keep those errors. I replaced it with `key_exists`, which keeps this distinction.

lwjohnst86 (approved):
looooksss goodd :grin:

PR #218, opened 2025-11-26
feat: :sparkles: check primary key exists

### LINKED ISSUE #216: Primary key fields must exist

martonvago:
# Description

This PR implements the MUST constraint that primary key fields must exist (i.e. they must be fields on the same resource).

Closes #216 

Needs an in-depth review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`

### DIFF
--- a/src/check_datapackage/check.py
+++ b/src/check_datapackage/check.py
@@ -150,11 +150,11 @@ def _check_keys(properties: dict[str, Any], issues: list[Issue]) -> list[Issue]:
     resources_with_pk = _keep_resources_with_no_issue_at_property(
         resources_with_pk, issues, "schema.primaryKey"
     )
-    issues = _flat_map(resources_with_pk, _check_primary_key)
+    key_issues = _flat_map(resources_with_pk, _check_primary_key)
 
     # Foreign keys
 
-    return issues
+    return key_issues
 
 
 def _issues_at_property(
--- a/src/check_datapackage/check.py
+++ b/src/check_datapackage/check.py
@@ -172,8 +172,9 @@ def _only_resources_with_correct_property_at(
 
 def _check_primary_key(resource: PropertyField) -> list[Issue]:
     """Check that primary key fields exist in the resource."""
-    pk_fields = _key_fields_as_str_list(resolve("/schema/primaryKey", resource.value))
-    unknown_fields = _get_unknown_key_fields(pk_fields, resource.value)
+    pk_fields = resolve("/schema/primaryKey", resource.value)
+    pk_fields_list = _key_fields_as_str_list(pk_fields)
+    unknown_fields = _get_unknown_key_fields(pk_fields_list, resource.value)
 
     if not unknown_fields:
         return []
@@ -185,6 +186,7 @@ def _check_primary_key(resource: PropertyField) -> list[Issue]:
             message=(
                 f"No fields found in resource for primary key fields: {unknown_fields}."
             ),
+            instance=pk_fields,
         )
     ]
 
--- a/src/check_datapackage/check.py
+++ b/src/check_datapackage/check.py
@@ -129,7 +129,7 @@ def check(
         _set_should_fields_to_required(schema)
 
     issues = _check_object_against_json_schema(properties, schema)
-    issues += _check_keys(properties, _map(issues, lambda issue: issue.jsonpath))
+    issues += _check_keys(properties, issues)
     issues += apply_extensions(properties, config.extensions)
     issues = exclude(issues, config.exclusions, properties)
     issues = sorted(set(issues))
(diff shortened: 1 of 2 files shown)


src/check_datapackage/check.py
```diff
@@ -97,6 +101,78 @@ class for more details, especially about the default values.
...
+def _check_keys(properties: dict[str, Any], issue_jsonpaths: list[str]) -> list[Issue]:
+    """Check that primary and foreign keys exist."""
+    # Primary keys
+    resources_with_pk = _get_fields_at_jsonpath(
+        "$.resources[?(length(@.schema.primaryKey) > 0 || @.schema.primaryKey == '')]",
+        properties,
+    )
+    resources_with_pk = _only_resources_with_correct_property_at(
+        "schema.primaryKey", resources_with_pk, issue_jsonpaths
+    )
```
martonvago: I'm filtering out any resources that have something wrong with the PK because I don't want to check a PK that is broken in some more fundamental way (e.g. is of the wrong type).

src/check_datapackage/check.py
```diff
@@ -97,6 +101,78 @@ class for more details, especially about the default values.
     return issues
 
 
+def _check_keys(properties: dict[str, Any], issue_jsonpaths: list[str]) -> list[Issue]:
+    """Check that primary and foreign keys exist."""
+    # Primary keys
+    resources_with_pk = _get_fields_at_jsonpath(
+        "$.resources[?(length(@.schema.primaryKey) > 0 || @.schema.primaryKey == '')]",
+        properties,
+    )
```
martonvago: Using `jsonpath` to query the properties is super neat! Much easier than indexing into it in straight Python.

src/check_datapackage/check.py
```diff
@@ -97,6 +101,78 @@ class for more details, especially about the default values.
...
+
+    if not unknown_fields:
+        return []
+
+    return [
+        Issue(
+            jsonpath=f"{resource.jsonpath}.schema.primaryKey",
+            type="primary-key",
+            message=(
+                f"No fields found in resource for primary key fields: {unknown_fields}."
```
martonvago: This lists all unknown fields in the same issue. Alternatively, we could have a separate issue for each.

src/check_datapackage/check.py
```diff
@@ -97,6 +101,78 @@ class for more details, especially about the default values.
     return issues
 
 
+def _check_keys(properties: dict[str, Any], issue_jsonpaths: list[str]) -> list[Issue]:
```
```suggestion
def _check_keys(properties: dict[str, Any], issues: list[Issue]) -> list[Issue]:
```

src/check_datapackage/check.py
```diff
@@ -97,6 +101,78 @@ class for more details, especially about the default values.
...
+    """Check that primary and foreign keys exist."""
+    # Primary keys
+    resources_with_pk = _get_fields_at_jsonpath(
+        "$.resources[?(length(@.schema.primaryKey) > 0 || @.schema.primaryKey == '')]",
+        properties,
+    )
+    resources_with_pk = _only_resources_with_correct_property_at(
+        "schema.primaryKey", resources_with_pk, issue_jsonpaths
```
```suggestion
        resources_with_pk, jsonpaths, "schema.primaryKey"
```

Revise ordering so main objects we are working with are first in the position.

src/check_datapackage/check.py
```diff
@@ -136,6 +140,80 @@ class for more details, especially about the default values.
...
+
+    # Foreign keys
+
+    return issues
+
+
+def _only_resources_with_correct_property_at(
+    jsonpath: str, resources: list[PropertyField], issue_jsonpaths: list[str]
+) -> list[PropertyField]:
+    """Filter out resources that have an error at or under the given `jsonpath`."""
```
```suggestion
    """Filter out resources that have an issue at or under the given `jsonpath`."""
```

?

lwjohnst86 (changes requested):
Nice! Some comments and suggested edits :relaxed:

src/check_datapackage/check.py
martonvago: Just to say that if you suggest a change (like changing the arg to `issues` from `issue_jsonpaths`), it's a lot of extra work for you to make that suggestion everywhere in the PR. If it's easier, you can just suggest it in one place and I will know to update the code accordingly everywhere else.

lwjohnst86: Thanks! Yea, in this case, it helped me to review and work through what the code was doing, so it was a good use of my time :stuck_out_tongue_closed_eyes:

martonvago:
Oops, I think `_keep_resources_with_no_issue_at_property` is a bit broken now, moving this to in progress while I look into it

lwjohnst86 (approved):

Issue #211, opened 2025-11-25
Brainstorming style and aesthetics of `explain()` messages


lwjohnst86:
> Personally, I think it can look a bit cluttered with raw markup being printed as output. Could we punt this to a PR that deals with the styles of these messages and considers using rich? E.g. just using the `print` from rich would give us some additional formatting:

<img width="450" height="274" alt="image" src="https://github.com/user-attachments/assets/8c51d308-836a-46dd-b467-30173b001529" />

_Originally posted by @joelostblom in https://github.com/seedcase-project/check-datapackage/pull/208#discussion_r2559113205_

joelostblom:
After #244 and #246, error message would look like this:

<img width="770" height="262" alt="Image" src="https://github.com/user-attachments/assets/c8d28567-fce7-4c97-b06d-04c1bade39c7" />

@lwjohnst86 I remember you mentioned before that you maybe wanted the JSON paths formatted somehow as well. Do you still think so and if so which of [rich's formatting styles/colors](https://rich.readthedocs.io/en/latest/style.html#defining-styles) do you think we should apply?

Here is couple of examples for the most widely supported options:

<img width="770" height="262" alt="Image" src="https://github.com/user-attachments/assets/119138db-7d6f-4911-bf14-53ac2475886b" />


Bold (a bit more subtle):
<img width="770" height="262" alt="Image" src="https://github.com/user-attachments/assets/a243bbfe-77fb-4a03-a138-b3c3bd56502b" />

PR #241, opened 2025-12-11
fix: :bug: remove package prefix in error message

### LINKED ISSUE #240: Remove hard-coded "package" prefix in `explain()`
https://github.com/seedcase-project/check-datapackage/blob/5f2bc0615514c01dd05b4c1fe05e639d0eb17297/src/check_datapackage/check.py#L98

Should not be hard coded as `package`.

So maybe `At '{jsonpath}':`. 

*Follow up from https://github.com/seedcase-project/seedcase-sprout/pull/1605#discussion_r2598081227*

joelostblom:
# Description

The hardcoded prefix led to some issues when integrating check-datapackage into sprout. See #240 for details.

Test with:

```sh
uv run python -c 'import check_datapackage as cdp; cdp.check({"resources": []}, error=True)'
```

Output before this PR:

```
1 issue was found in your `datapackage.json`:

At package.resources:
|
| resources: []
|            ^^
[] should be non-empty
```

Output after this PR:

```
1 issue was found in your `datapackage.json`:

At resources:
|
| resources: []
|            ^^
[] should be non-empty
```

I will say that with a top level path like this, it does appear a bit redundant to list `resources` twice and the `package.resource` aesthetically gave a bit more variation, but it is minor and for more nested paths this will look good and be shorter.

Closes #240

Needs a quick review.

## Checklist

- [ ] Formatted Markdown
- [x] Ran `just run-all`


src/check_datapackage/check.py
```diff
@@ -95,7 +95,7 @@ def _create_explanation(issue: Issue) -> str:
     property_name = issue.jsonpath.removesuffix("$").split(".")[-1]
     number_of_carets = len(str(issue.instance))
     return (  # noqa: F401
-        f"At package{issue.jsonpath.removeprefix('$')}:\n"
+        f"At {issue.jsonpath.removeprefix('$').removeprefix('.')}:\n"
```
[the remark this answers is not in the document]
martonvago: Yeah, I think it always starts with `$.`. Except when the instance is the root node itself (`$`), but you're dealing with that separately in #243 👍

martonvago (approved):

src/check_datapackage/check.py
```diff
@@ -95,7 +95,7 @@ def _create_explanation(issue: Issue) -> str:
     property_name = issue.jsonpath.removesuffix("$").split(".")[-1]
     number_of_carets = len(str(issue.instance))
     return (  # noqa: F401
-        f"At package{issue.jsonpath.removeprefix('$')}:\n"
+        f"At {issue.jsonpath.removeprefix('$.')}:\n"
```
```suggestion
        f"At `{issue.jsonpath.removeprefix('$.')}`:\n"
```

very small thing, but it helps to have something that visually indicates what this is or isn't.

lwjohnst86 (changes requested):
very nice! Just a small change

src/check_datapackage/check.py
joelostblom: I made a comment about possible rich styles for the JSON path in
https://github.com/seedcase-project/check-datapackage/issues/211#issuecomment-3645568749. I think it is a separate concept that could be part of another PR?

I also opened https://github.com/seedcase-project/check-datapackage/issues/250 to discuss the need for fallback markup.

lwjohnst86 (approved):

PR #246, opened 2025-12-12
feat: :sparkles: include error type and add red color via `rich`

joelostblom:
*Rebase and merge after https://github.com/seedcase-project/check-datapackage/pull/244*

# Description

I noticed that we were suppressing printing the error type in our custom exception hook for DataPackage errors. I think it would be useful to see the type of error so I added that here. While doing that, I thought it could also help to colorize this error the same as the carets, to help visually make the connection that the carets are pointing to where in the code the error occurred.

Input:

```zsh
uv run python -c 'import check_datapackage as cdp; cdp.check({"resources": [{"title": "Title", "name": 123}]}, error=True)'
```

Output before this PR (after #244):
image

Output after this PR:
image

Potentially closes https://github.com/seedcase-project/check-datapackage/issues/211 unless we want to keep that open for further discussion.

Needs a quick review.

## Checklist

- [ ] Formatted Markdown
- [x] Ran `just run-all`

PR #244, opened 2025-12-11
feat: :sparkles: colorize output of `explain()` via `rich`

joelostblom:
# Description

Rich's print function automatically colors different types of of variables and some markup. A first step towards https://github.com/seedcase-project/check-datapackage/issues/211, where more specific improvements can be discussed further.

Needs a quick review.

## Checklist

- [ ] Formatted Markdown
- [x] Ran `just run-all`

PR #249, opened 2025-12-12
fix: :bug: no traceback and correct rich formatting in IPython

### LINKED ISSUE #248: IPython/Jupyter does not render rich output correctly
In addition to the traceback not being suppressed as per https://github.com/seedcase-project/seedcase-sprout/issues/1608, IPython does not render the rich markup styles correctly and just shows the markup tag instead (`[red]` etc):

Image

### LINKED ISSUE #1608: IPython Notebooks output the traceback even though we turn it off
In the [guide](https://check-datapackage.seedcase-project.org/docs/guide/check#stop-program-on-failed-checks-errortrue) it doesn't turn off even though it is off outside of IPython Notebook.

joelostblom:
*Rebase and merge after #244 and #246*

# Description

IPython has its own exception handler and does not respect `sys.excepthook`. The solution here is based on [this SO answer](https://stackoverflow.com/a/70433500/2166823) and [this code snippet](https://gist.github.com/waszil/fccda8749ed5dd5baa336bae8e88dd7a). Note that [the IPython docs](https://ipython.readthedocs.io/en/stable/api/generated/IPython.core.interactiveshell.html#IPython.core.interactiveshell.InteractiveShell.set_custom_exc) include a warning about modifying the traceback (I don't know if there is actually any additional risk from modifying the system traceback as we are currently doing):

> WARNING: by putting in your own exception handler into IPython’s main execution loop, you run a very good chance of nasty crashes. This facility should only be used if you really know what you are doing.

Test in IPython/Jupyter with:

```py
import check_datapackage as cdp; cdp.check(123, error=True)
```

### Output before this PR:
image

### Output after this PR:
image

### Docs before this PR:
image

### Docs after this PR:
image

Closes #248, closes seedcase-project/seedcase-sprout#1608

---

Needs an in-depth review.

## Checklist

- [ ] Formatted Markdown
- [x] Ran `just run-all`

### DIFF
--- a/src/check_datapackage/check.py
+++ b/src/check_datapackage/check.py
@@ -85,7 +85,7 @@ def explain(issues: list[Issue]) -> str:
     num_issues = len(issue_explanations)
     singular_or_plural = " was" if num_issues == 1 else "s were"
     return (
-        f"{num_issues} issue{singular_or_plural} found in your [u]datapackage.json[/u]:\n\n"
+        f"{num_issues} issue{singular_or_plural} found in your [u]datapackage.json[/u]:\n\n"  # noqa: E501
         + "\n".join(issue_explanations)
     )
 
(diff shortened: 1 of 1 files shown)


lwjohnst86 (approved):
Just a comment, but approved!

PR #251, opened 2025-12-15
fix: :bug: add rich as dependency

martonvago:
# Description

We should declare `rich` as a dependency. (Was probably working anyway because of a transitive dependency, but this won't always work in projects that use `cdp` or Sprout?)

Needs a quick review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`


lwjohnst86 (approved):
Nice catch :+1:

PR #253, opened 2025-12-16
fix: :bug: `properties` argument wasn't actually used, so removed

lwjohnst86:
# Description

This argument wasn't even used, so just removed it.

Needs a quick review.

## Checklist

- [x] Ran `just run-all`


martonvago (approved):
🎉

PR #259, opened 2025-12-18
docs: :lipstick: use `rich` for pretty output in guide docs

### LINKED ISSUE #238: Use rich print when showing properties/dict in guide
It formats it better than pprint.

joelostblom:
# Description

Needs a quick review
Closes #238

Make output easier to read, esp larger objects like dictionaries. This adds both coloring and formatting. See commit messages for more details.

Currently, output like multiple issues is a mess and hard to read:
image

After this PR the structure is much better (the default colors are a bit off for a light bg; will be dealt with in a follow up PR:
image

Before (soft-wrapping and no color):
Screenshot from 2025-12-18 08-31-25

After (no wrapping and color):
Screenshot from 2025-12-18 08-30-58

Sometimes the lack of wrapping becomes a bit of an issue, but this will be dealt with in a follow up PR.

Before:
Screenshot from 2025-12-18 08-21-27

After (would be nice if his wrapped or was spaced out more (in follow up PR):
Screenshot from 2025-12-18 08-21-33

Note that using the `__repr__` of the output is slightly different from pretty printing it; the differences are small and if anything I slightly prefer this approach.
Screenshot from 2025-12-16 18-16-02

## Checklist

- [x] Ran `just run-all`
    - I'm seeing some failure checking typos on a JS method in an HTML file, but that is probably not related to this PR... (maybe the built doc pages are not properly excluded from the typo check?)

### DIFF
--- a/docs/guide/check.qmd
+++ b/docs/guide/check.qmd
@@ -5,6 +5,8 @@ jupyter: python3
 order: 1
 ---
 
+{{< include ../includes/_pretty_repr.qmd >}}
+
 The purpose of `check-datapackage` is to make sure a Data Package's
 metadata---stored in its `datapackage.json` file---complies with the
 [Data Package standard](https://datapackage.org). This standard defines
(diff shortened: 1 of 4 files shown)


martonvago (approved):
I think this is a massive improvement and looks really nice! 🪄

lwjohnst86 (approved):
Very nice :tada:

PR #260, opened 2025-12-18
docs: :lipstick: improve code output format and color in docs

joelostblom:
# Description

Needs a quick review.

As mentioned in #259, there are a few issues with the default rich representation of the output: the defaults colors don't look that great on a white background, the lack of soft-wrap is sometimes and issue, and there is also an extra newline inserted after each code block before the output appears. These issues can be seen in this example screenshot showing an `Issue` instance:
image

After this PR, all three issues are fixed and the output is now styled much closer to the appearance of the default quarto theme for the input code blocks:
Screenshot from 2025-12-18 10-05-34

## Checklist

- [x] Ran `just run-all`

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -66,3 +66,6 @@ dev = [
     "typos>=1.35.6",
     "vulture>=2.14",
 ]
+
+[tool.uv.sources]
+rich = { git = "https://github.com/joelostblom/rich", rev = "ipython-console-param" }
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,6 +1,6 @@
 [project]
 name = "check-datapackage"
-version = "0.27.0"
+version = "0.28.0"
 description = "Check your Data Package's metadata against the Data Package standard to ensure that it's compliant with the standard"
 authors = [
   {name = "Luke W. Johnston", email = "lwjohnst@gmail.com" },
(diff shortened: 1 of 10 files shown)


lwjohnst86 (approved):
Lovely :tada:

pyproject.toml
```diff
@@ -66,3 +66,6 @@ dev = [
     "typos>=1.35.6",
     "vulture>=2.14",
 ]
+
+[tool.uv.sources]
+rich = { git = "https://github.com/joelostblom/rich", rev = "ipython-console-param" }
```
joelostblom: I'm just adding a note here that this [only affects the version of rich installed as a developmental dependency](https://docs.astral.sh/uv/concepts/projects/dependencies/#dependency-sources#:~:text=which%20are%20used%20during%20development) (ie when running `uv add`), not when the package is installed for production. I think this is desired since we only the forked functionality for docs and ideally don't want to run a custom fork for production.

Even more ideal would be if it only was installed for the doc build. I didn't find a clear way to do that via `pyproject.toml`, but it might be possible if we dig a bit deeper. It might also be possible to achieve this by changing the doc build command to include the fork in a `--with` flag.

In either case, I don't think the changes I did in the fork are big enough to spend more time on this now, particularly since production installations are not affected, but thought I would note this.

For the future, these were the commands I used to check which installation of rich was used (and then I manually checked the file `pretty.py` in that location to see whether it had the changes I made). Both commands showed that the pypi version of rich was installed, not my fork:


```sh
uv run --with  "git+https://github.com/seedcase-project/check-datapackage" --no-project -- python -c "import rich; print(rich.__file__)"
```


```sh
python -m pip install "git+https://github.com/seedcase-project/check-datapackage"; python -c "import rich; print(rich.__file__)"
```

PR #261, opened 2025-12-19
build: :hammer: add doc preview action to build

joelostblom:
# Description

Same change as in the other repos a while ago.

Needs a quick review.

## Checklist

- [x] Ran `just run-all`


lwjohnst86 (approved):
Nice :grin:

Issue #263, opened 2025-12-20
Output from `explain(issues)` is currently not that helpful when run on its own without `check(..., error=True)`


joelostblom:
If I currently run `explain` on a set of issues, the return is not that useful since it is a string with markup for formatting:

```py
import check_datapackage as cdp
package_properties = cdp.example_package_properties()
del package_properties["resources"][0]["path"]
issues = cdp.check(properties=package_properties)
cdp.explain(issues)
```

<img width="866" height="244" alt="Image" src="https://github.com/user-attachments/assets/558e31f5-bb7a-4e61-92ce-8ef36ad6e002" />

If we want explain to be used this way, I think we need to rethink the output from it both with and without `error=True`. A few guiding questions:

1. Do we want `explain` to be a public function runnable from code by the user or a helper function to nicely format `DataPackageError`?
    - Currently, we have hardcoded the path `datapackage.json` in the `explain` output, that would need to change if explain is a public function that is run on other input than this json file.
2. If explain is a public function, should it return a string that the user can do something with or just print a helpful message.
3. If explain is a public function runnable from code, then how should it interface with `DataPackageError`? 
   - E.g. if we add a print directly in explain and then run `check(..., error=True)`, all the printed output from explain will appear above the line that says `DataPackageError`, something like this:
        <img width="866" height="400" alt="Image" src="https://github.com/user-attachments/assets/01e4d7c9-42b3-49d0-bef6-8c307c64081f" />

Tagging @martonvago and @lwjohnst86 since we worked on the `explain` implementation.

PR #264, opened 2025-12-20
docs: :memo: run `explain()` code chunk

joelostblom:
# Description

We should run this code chunk now that explain works. However, currently the output is unhelpful as per https://github.com/seedcase-project/check-datapackage/issues/263, so we might want to resolve that before merging this. If we decide there that the output from explain should just be a string, then maybe we want to show how to pretty print it manually here?

Needs a quick review.

## Checklist

- [x] Ran `just run-all`


lwjohnst86 (approved):
In this case, we can use rich print to make the output prettier. I added it to the PR :)

PR #309, opened 2026-03-11
feat: ✨ generalize beautification of errors

joelostblom:
# Description

It is helpful to remove the traceback for more errors than just the datapackage errors. Making this function more general and adding it as part of the public interface makes it usable from e.g. flower. Long term, it maybe makes sense to move this (together with `read_properties` etc), to a `seedcase-utils` package? Might be more convenient than creating a different pypi package for each smaller util we make.

Needs an in-depth review.

## Checklist

- [x] Formatted Markdown
- [x] Ran `just run-all`


src/check_datapackage/check.py
```diff
@@ -28,33 +28,89 @@
...
-    exc_type: type[BaseException],
-    exc_value: BaseException,
-    exc_traceback: TracebackType | None,
-) -> None:
-    """Exception hook to hide tracebacks for DataPackageError."""
-    if issubclass(exc_type, DataPackageError):
-        _pretty_print_exception(exc_type, exc_value)
-    else:
-        sys.__excepthook__(exc_type, exc_value, exc_traceback)
+def _create_suppressed_traceback_hook(
```
joelostblom: I renamed some of these functions for clarity

joelostblom (commented):
Ready for review!

@lwjohnst86 I agree we shouldn't change packages to accommodate other package's need for the same functionality. However, I also think we should avoid having the same code in two places as much as possible. In this case, it is actually not possible to simply copy the code over since in its current state two package's using this functionality would cause conflicts; more details in https://github.com/seedcase-project/check-datapackage/pull/309#discussion_r2943523477

I wonder if creating a utility package within seedcase would solve these types of issues so I initiated a discussion in https://github.com/seedcase-project/design/issues/245

lwjohnst86 (changes requested):
Phew, those tests didn't look fun to make, ugh, reading them wasn't fun at least 😝 Only small comments. I can set up the soil package so you can move that over that once this PR has been merged.

joelostblom (commented):
Yeah, I think tests like this is a good use of getting help from genAI. It works well to get started even if there is a lot of tweaking and even more reading to remove tests from the initial suggestions they make.

lwjohnst86 (approved):
:tada: made an issue to deal with improving this situation in a way that doesn't impact other packages that use this.

PR #316, opened 2026-03-31
docs: :pencil2: restructure to match Flower's design structure

joelostblom:
# Description

Before adding the CLI page, I thought it would be helpful if the docs structure is consistent with how we have it in flower so I restructured the pages accordingly. I tried to avoid adding new content as much as possible in this PR and just change the organization into multiple page.

Needs a thorough review.

## Checklist

- [x] Ran `just run-all`


docs/design/interface/python.qmd
```diff
@@ -1,49 +1,42 @@
 ---
-title: "Interface"
+title: Python
 ---
```
lwjohnst86, line 4:
```suggestion

This section of the documentation focuses on the design of the interface
and how it is implemented as Python code.

```

Just to keep this intro

docs/design/interface/config.qmd
```diff
@@ -0,0 +1,85 @@
...
+    """Reads a configuration file into a `Config` object.
+
+    Args:
+        path: The path to the configuration file.
+
+    Returns:
+        A `Config` object populated with the contents of the configuration
+            file.
+    """
+```
```
lwjohnst86: Since this is a function, it should be kept in the Python file along with the other functions.

_quarto.yml
```diff
@@ -38,11 +38,18 @@ website:
...
+        - text: "Architecture"
+          href: docs/design/architecture.qmd
+        - section: "Interface"
+          href: docs/design/interface/index.qmd
           contents:
-            - docs/design/architecture.qmd
-            - docs/design/interface.qmd
+            - docs/design/interface/inputs.qmd
+            - docs/design/interface/outputs.qmd
+            - docs/design/interface/cli.qmd
```
lwjohnst86: This will fail to build because this doesn't exist.


```suggestion
```

lwjohnst86 (changes requested):
Nice! Very minor things

docs/design/interface/config.qmd
joelostblom: Could you clarify, do you just want to move the function but keep the class here?

I was initially going to put both under the section of the Python interface, since you can interact with both of them via Python, but then I noticed you kept the config related classes on [the configuration page for flower](https://flower.seedcase-project.org/docs/design/interface/config#python-classes), so I kept both the class and function here.

What should my mental model be for this?

joelostblom (commented):
Thanks for the quick review @lwjohnst86! Accepted your suggestions, just one question.

docs/design/interface/config.qmd
lwjohnst86: That's a very good question. Hmm, I think we should update Flower so that all Python functions and classes are in the `python.qmd`, and follow that "heuristic" here too.

joelostblom: Sounds good! I removed it from this page and opened an issue in flower.

lwjohnst86 (approved):

PR #323, opened 2026-04-02
feat: :sparkles: implement CLI

### LINKED ISSUE #307: Implement a `check_cli()` function for CLI
I'm not sure if this is necessary, but it won't be good to have a name clash with `check()`. Maybe call it `check_cli()`, but set the CLI command name to be `check`?

joelostblom:
# Description

This is an initial implementation of what the cli of cdp could look like. This is intentionally kept minimal and trying to use as many of the existing cdp functions as possible before adding the more elaborate ones that were developed in flower (e.g. read_json instead of read_source) to not convolute the addition of the CLI with the addition of new capabilities.

Depends on the functionality being merged into soil and there being a first release of that package (can be tested as is, but we should ideally not point straight to an unversion git repo as a dependency)

Review with https://github.com/seedcase-project/seedcase-soil/pull/5/

Closes #307 

Needs a thorough review.

## Checklist

- [x] Ran `just run-all`


martonvago (dismissed):
Looks really nice! 🎉

src/check_datapackage/cli.py
```diff
@@ -0,0 +1,49 @@
...
+    """Check a Data Package's metadata against the Data Package standard.
+
+    Outputs a human-readable explanation of any issues found.
+
+    Args:
+        source: The local location of a `datapackage.json` file.
+        strict: If True, check "SHOULD" properties in addition to "MUST"
+            properties from the Data Package standard.
+    """
+    properties: dict[str, Any] = read_json(Path(source))
```
lwjohnst86, line 42: Related to your comment about the `read_source()`, in `check-datapackage`, we need read in without assuming anything, unlike with Flower that uses a basic check of the input metadata. So reading in can't also check, as that would be circular 😝

src/check_datapackage/cli.py
```diff
@@ -0,0 +1,49 @@
...
+        "standard"
+    ),
+    config_name=".cdp.toml",
+)
+
+
+@app.command(name="check")
+def check_cmd(
+    source: str = "datapackage.json",
+    *,
```
lwjohnst86: What's this for?

src/check_datapackage/cli.py
```diff
@@ -0,0 +1,49 @@
...
+    strict: bool = False,
+) -> None:
+    """Check a Data Package's metadata against the Data Package standard.
+
+    Outputs a human-readable explanation of any issues found.
+
+    Args:
+        source: The local location of a `datapackage.json` file.
```
lwjohnst86, line 38:
```suggestion
        source: The location of a `datapackage.json` file.
```

Doesn't strictly have to be local.

lwjohnst86 (changes requested):
Great start! Can you start including some integration tests here?

src/check_datapackage/cli.py
joelostblom: I opted to describe the current functionality of code. Will update to match the design docs when we actually add the capabilities for reading remote files.

src/check_datapackage/cli.py
joelostblom: Yes, agree! My plan is to refactor out `check` from `read_properties` when we hoist it to soil. Then it becomes an expclicit step in flower instead, which also better matches the naming of that functionality. So in check_datapackage it will be

```py
address: Address = parse_source(source)
properties: dict[str, Any] = read_properties(address)
```

hereas in flower it will be

```py
address: Address = parse_source(source)
properties: dict[str, Any] = read_properties(address)
check(datapackage, error=True)
```

src/check_datapackage/cli.py
joelostblom: It's to differentiate positional args from keyword params. I removed it from this PR and opened https://github.com/seedcase-project/seedcase-flower/issues/264 to discuss instead.

joelostblom (commented):
@lwjohnst86  Updated based on your comments and added tests!

src/check_datapackage/cli.py
lwjohnst86: I like, great plan :+1: :grin:

src/check_datapackage/cli.py
lwjohnst86: OOoooo :eyes: that's neat!

src/check_datapackage/cli.py
```diff
@@ -0,0 +1,48 @@
...
+        "check-datapackage checks if metadata is compliant with the Data Package"
+        "standard"
+    ),
+    config_name=".cdp.toml",
+)
+
+
+@app.command(name="check")
+def check_cmd(
+    source: str = "datapackage.json",
```
lwjohnst86, line 28: You can bring back that separator! Between positional and keyword. I think that's a nice way to clearly indicate which is the most important parameter there 👍

tests/test_cli.py
```diff
@@ -0,0 +1,84 @@
...
+    assert bound.arguments["strict"] is True
+
+
+# Success and error handling ====
+
+
+def test_check_valid_datapackage_succeeds(
+    capsys, datapackage_path, tmp_path, monkeypatch
+):
+    """A valid datapackage should pass checks."""
```
lwjohnst86:
```suggestion
    """A correct data package should pass checks."""
```

tests/test_cli.py
```diff
@@ -0,0 +1,84 @@
...
+    with pytest.raises(FileNotFoundError):
+        app(["check", "nonexistent.json"], result_action="return_value")
+
+
+def test_check_raises_error_on_validation_failure(tmp_path, monkeypatch):
+    """Check should raise DataPackageError when validation fails."""
+
+    invalid_datapackage = {"name": "invalid-name!"}
+    file_path = tmp_path / "datapackage.json"
+    file_path.write_text(json.dumps(invalid_datapackage))
```
lwjohnst86:
```suggestion
def test_check_raises_error_on_failure(tmp_path, monkeypatch):
    """Failed check should give DataPackageError."""

    wrong_datapackage = {"name": "bad-name!"}
    file_path = tmp_path / "datapackage.json"
    file_path.write_text(json.dumps(wrong_datapackage))
```

Given the design text we have about the word valid 😉

tests/test_cli.py
```diff
@@ -0,0 +1,84 @@
...
+
+@pytest.fixture
+def mock_read_json(mocker):
+    """Mock read_json to isolate CLI tests from filesystem resolution."""
+    return mocker.patch("check_datapackage.cli.read_json")
+
+
+@pytest.fixture
+def mock_check(mocker):
+    """Mock check to isolate CLI from validation logic."""
```
lwjohnst86:
```suggestion
    """Mock check to isolate CLI from logic."""
```

tests/conftest.py
```diff
@@ -1,5 +1,40 @@
+import json
 import os
 
+import pytest
+
+
+@pytest.fixture
+def datapackage():
+    """Return a valid datapackage dict with resources."""
```
lwjohnst86:
```suggestion
    """Returns a data package dict with resources."""
```

lwjohnst86 (changes requested):
Very minor changes ☺️

src/check_datapackage/cli.py
```diff
@@ -0,0 +1,50 @@
...
+    /,
+    *,
```
lwjohnst86:
```suggestion
    /, # This means that the above are positional parameters for the CLI
    *, # This means that the below are keyword parameters for the CLI
```

Just to document what those mean. I don't think they are a commonly used feature.

lwjohnst86 (changes requested):
Very minor last thing!

src/check_datapackage/cli.py
joelostblom: I added a shorter comment to indicate this. Note that it's not only for the CLI, when calling the function via the python interface it works the same way.

src/check_datapackage/cli.py
```diff
@@ -0,0 +1,50 @@
...
+
+    Args:
+        source: The local location of a `datapackage.json` file.
+        strict: If True, check "SHOULD" properties in addition to "MUST"
+            properties from the Data Package standard.
+    """
+    properties: dict[str, Any] = read_json(Path(source))
+    config = Config(strict=strict)
+    check(properties, config=config, error=True)
+    pretty_print("[green]All checks passed! Your Data Package is valid.[/green]")
```
lwjohnst86:
```suggestion
    pretty_print("[green]All checks passed![/green]")
```

Not sure we need that last bit.

lwjohnst86 (changes requested):
One last small thing otherwise 👍

src/check_datapackage/cli.py
joelostblom: I agree it's better without it! Updated and ready for another look!

lwjohnst86 (approved):
:tada: greaatttt :+1:

PR #329, opened 2026-04-17
docs: 📝 Use more specific properties class

### LINKED ISSUE #13: `read_properties` could return `PackageProperties` instead of `dict`
Is there any particular reason we didn't use our custom class here? This change would make it more self explanatory, but it would involve hoisting `properties.py` from sprout into soil (which seems to make sense to me, but I'm noting it since we are now migrating quite a lot of key pieces here).

joelostblom:
# Description

Closes https://github.com/seedcase-project/seedcase-soil/issues/13

Needs thorough review.

## Checklist

- [x] Ran `just run-all`


lwjohnst86 (approved):
:tada: easy peasy :grin:

PR #332, opened 2026-04-21
docs: 📝 add guide on using the CLI

joelostblom:
# Description

Closes #306. Merge after https://github.com/seedcase-project/seedcase-soil/pull/25

Needs a thorough review.

## Checklist

- [x] Ran `just run-all`

PR #333, opened 2026-04-22
chore: 🔧 update soil so that docs build correctly

joelostblom:
# Description

Follow up on https://github.com/seedcase-project/check-datapackage/pull/332 where the soil version was not updated after updating the syntax in the docs to use the new examples.

Needs a quick review.

## Checklist

- [x] Ran `just run-all`


pyproject.toml
```diff
@@ -74,4 +74,4 @@ dev = [
 ]
 
 [tool.uv.sources]
-seedcase-soil = { git = "https://github.com/seedcase-project/seedcase-soil", rev = "0.4.0" }
+seedcase-soil = { git = "https://github.com/seedcase-project/seedcase-soil", rev = "0.7.1" }
```
lwjohnst86: Soil is on PyPI so this can be removed ☺️

lwjohnst86 (changes requested):

pyproject.toml
joelostblom: Silly me, updated.

lwjohnst86 (changes requested):
Should I wait on this until after the IO has been merged in?

docs/guide/cli.qmd
```diff
@@ -46,15 +46,9 @@ Package, `flora`.
...
+import seedcase_soil as so
 
-from seedcase_soil import ExampleDataPackageName, write_example_datapackage
-
-datapackage_path = write_example_datapackage(
-    Path("."),
-    ExampleDataPackageName.flora,
-)
-!uv run check-datapackage check {datapackage_path}
+!uv run check-datapackage check {so.Example.flora.address.value}
```
joelostblom, line 51: I thought of this much less verbose syntax to read the path from the CLI in the docs examples.

joelostblom (commented):
Yup, I thought this would be a quick "critical" fix to get the site live and that we then updated it again after we merged the example PR. But now that we merged the example PR first, it's great that you waited! I updated this PR accordingly.

lwjohnst86 (approved):

PR #334, opened 2026-04-23
docs: 📝 add configuration guide

### LINKED ISSUE #324: Add "creating/using config file" guide

joelostblom:
# Description

Follows the same format as Flower.

Closes #324 

Needs a quick review.

## Checklist

- [x] Ran `just run-all`


docs/guide/config-file.qmd
```diff
@@ -0,0 +1,72 @@
...
+
+`check-datapackage` can be configured through command-line options and a
+configuration file. This guide covers configuration files for the CLI.
+For command-line usage, see the [CLI guide](/docs/guide/cli.qmd). For
+the Python `Config` class used with `check()`, see
+[Configuring the checks](/docs/guide/config.qmd).
+
+## Configuration files
+
+A configuration file lets you set preferences once, so you do'nt have to
```
```suggestion
A configuration file lets you set preferences once, so you don't have to
```

docs/guide/config-file.qmd
```diff
@@ -0,0 +1,72 @@
...
+repeat them every time you run `check-datapackage check`.
+`check-datapackage` supports two configuration files: `.cdp.toml`
+`pyproject.toml`. These files should be placed in the root of your
+project—`check-datapackage` will find them automatically.
```
```suggestion
project---`check-datapackage` will find them automatically.
```

docs/guide/config-file.qmd
```diff
@@ -0,0 +1,72 @@
...
+:::
+
+### Configurable settings
+
+Before diving into editing your configuration, let’s see which settings
+can be configured via the configuration file:
+
+- `strict`: If `true`, include checks for Data Package properties marked
+  as "SHOULD" in addition to those marked as "MUST". Defaults to
+  `false`.
```
lwjohnst86, line 38: Ummm, there's a lot more things you can configure here. E.g. see https://check-datapackage.seedcase-project.org/docs/design/interface/config#overview Are those not implemented yet in the CLI?

lwjohnst86 (changes requested):
Good start, but there's a lot more things to configure in the file.

docs/guide/config-file.qmd
joelostblom: Nope, we did a minimal implementation for the CLI in https://github.com/seedcase-project/check-datapackage/pull/323 to only support the simple options such as `strict`. It is unclear to me if we should expose the others via the CLI since they cant take multiline input, which we opted not to expose in the CLI for flower and instead only allow for being read from a config file. There, we even moved them to their own config file, so we never encountered the issue we are now seeing here. Maybe a good solution could be to introduce params that are hidden from the CLI but still read from the config file? I can draft a PR for that including updated config docs, but maybe we can merge this PR in the meanwhile as documentation for the currently supported configuration and then build on it in follow up PRs?

lwjohnst86: Hmmmm, right right, I forgot that. We can hide params from the CLI? Yea, that works. Odd that the config can't take non-CLI arg options... :thinking: How do other tools do it then? :thinking:

joelostblom: Yes, I created and example of how to hide params in https://github.com/seedcase-project/check-datapackage/pull/340; they can either be hidden just from the help menu or not parsed at all.

I'm not sure how other tools using cyclopts does this, but it is possible to pass non-CLI config options to the Config class if we write some custom logic ourselves. It just seemed to me like the most cyclopts native way of doing this was to declare that the function takes some parameters that can only be read from the config file and not passed via the CLI (to make use of all the existing configuration handling in cyclopts)

lwjohnst86 (approved):

PR #340, opened 2026-04-28
feat: ✨ support exclusions via config file

### LINKED ISSUE #303: Test out that config file structure is read in correctly with cyclopts

joelostblom:
Merge after https://github.com/seedcase-project/check-datapackage/pull/334

# Description

Support settings exclusions via the config file, but not via the CLI. This is done creating a hidden CLI param so that cyclopts automatically detects to read this param from the config. 

Relevant comment from https://github.com/seedcase-project/check-datapackage/pull/334/changes#r3152515787:

>  It is unclear to me if we should expose all params via the CLI since they cant take multiline input, which we opted not to expose in the CLI for flower and instead only allow for being read from a config file. There, we even moved them to their own config file, so we never encountered the issue we are now seeing here. Maybe a good solution could be to introduce params that are hidden from the CLI but still read from the config file? 

Needs a thorough review.

Maybe also closes https://github.com/seedcase-project/check-datapackage/issues/303 since we have an example here of the config file being read correctly?

## Checklist

- [x] Ran `just run-all`


src/check_datapackage/cli.py
```diff
@@ -30,6 +32,7 @@ def check_cmd(
     /,  # End of positional-only args
     *,  # Start of keyword-only params
     strict: bool = False,
+    exclusions: Annotated[list[Exclusion], Parameter(show=False)] = [],
```
joelostblom, line 35: `show=False` hides a param from the help screen. There is also an option to use `parse=False` which makes cyclopts not parse it at all https://cyclopts.readthedocs.io/en/stable/default_parameter.html#skipping-private-parameters. That might be more suitable for what we want.

docs/guide/config-file.qmd
```diff
@@ -0,0 +1,137 @@
...
+
+### Configurable settings
+
+Before diving into editing your configuration, let’s see which settings
+can be configured via the configuration file:
+
+- `strict`: If `true`, include checks for Data Package properties marked
+  as "SHOULD" in addition to those marked as "MUST". Defaults to
+  `false`.
+- `exclusions`: Exclude issues by JSONPath, issue type, or both.
```
```suggestion
- `exclusions`: Exclude issues by [JSONPath](https://www.rfc-editor.org/rfc/rfc9535.html), issue type, or both.
```

src/check_datapackage/cli.py
lwjohnst86: Let's do `parse=False` instead. I'd rather not expose it at all.

lwjohnst86 (changes requested):
Lovely, just some suggested changes :)

src/check_datapackage/cli.py
joelostblom: I tried locally but it seems like "exclude from parsing" excludes it from the config file parsing as well, so `show` is the most suitable for our needs.

docs/guide/cli.qmd
```diff
@@ -36,48 +34,49 @@ Use the `check` command to check a Data Package's metadata:
...
-### Output
+```{python}
+#| echo: false
+so.format_output_for_docs(["cdp", "check", str(so.Example.simple.path)])
+```
 
-The example below shows the output of running `check-datapackage check`
-in a directory containing a `datapackage.json` for an example Data
-Package, `flora`.
+If a check fails, a `DataPackageError` is raised with more details about
```
lwjohnst86:
```suggestion
If a check fails, you'll get a `DataPackageError` with more details about
```

docs/guide/config-file.qmd
```diff
@@ -70,3 +71,74 @@ you used above in `.cdp.toml` below this table header.
 [tool.check-datapackage]
 strict = true
 ```
+
+## Exclusion examples
+
+The examples below use Soil's built-in `simple` Data Package:
```
```suggestion
The examples below use [Soil's](https://soil.seedcase-project.org) built-in [`simple` Data Package](https://soil.seedcase-project.org/docs/reference/example):
```

docs/guide/config-file.qmd
```diff
@@ -70,3 +71,74 @@ you used above in `.cdp.toml` below this table header.
...
+
+try:
+    so.format_output_for_docs(["cdp", "check", str(so.Example.simple.path)])
+finally:
+    config_path.unlink(missing_ok=True)
+```
+
+This fails because strict mode requires the `id` property.
+
+To exclude only that issue, add an exclusion that matches both the
```
```suggestion
To exclude only that issue, add an exclusion to the `.cdp.toml` file that matches both the
```

src/check_datapackage/cli.py
lwjohnst86: I was wondering about that. Makes sense that not parsing also doesn't allow the config to work. Thanks for testing it 😁

docs/guide/config-file.qmd
```diff
@@ -70,3 +71,74 @@ you used above in `.cdp.toml` below this table header.
...
+[[exclusions]]
+jsonpath = "$.id"
+type = "required"
+""")
+
+try:
+    so.format_output_for_docs(["cdp", "check", str(so.Example.simple.path)])
+finally:
+    config_path.unlink(missing_ok=True)
+```
```
lwjohnst86, line 144:
````suggestion
```

````

**Review by lwjohnst86 — CHANGES_REQUESTED** [review]
Nice! Some fairly small changes ☺️

File: docs/guide/config-file.qmd [hunk 7]
```diff
@@ -70,3 +71,73 @@ you used above in `.cdp.toml` below this table header.
...
+strict = true
+[[exclusions]]
+jsonpath = "$.id"
+type = "required"
+""")
+
+try:
+    so.format_output_for_docs(["cdp", "check", str(so.Example.simple.path)])
+finally:
+    config_path.unlink(missing_ok=True)
```
lwjohnst86, line 143:
````suggestion
    config_path.unlink(missing_ok=True)
```

````

Not sure what happened but it got mangled

lwjohnst86 (approved):

PR #341, opened 2026-04-28
build: 🔨 ensure docs don't pick up local config files

joelostblom:
# Description

Sometimes when I run the docs I forget that I have a local `cdp.toml` and don't understand why the output doesn't look as expected... I think this is a more general problem that could lead to issues with sublte differences between local build and the official pages that are hard to troubleshoot. This PR ensures that the docs run in an isolated environment that is resistant to being polluted by local configs.

I found this way of doing it and if it looks good we can port to the template, but I'm open to other solutions too. Open to other ideas and if someone sees something that could go wrong here; since we moved the examples out of the dirs I don't think we rely on reading from any local paths during code execution which would have been an issue otherwise.

Needs a thorough review.

## Checklist

- [x] Ran `just run-all`

### DIFF
--- a/docs/guide/cli.qmd
+++ b/docs/guide/cli.qmd
@@ -24,7 +24,7 @@ check-datapackage
 
 ```{python}
 #| echo: false
-!uv run check-datapackage
+!cdp
 ```
 :::
 
@@ -48,7 +48,7 @@ Package, `flora`.
 #| echo: false
 import seedcase_soil as so
 
-!uv run check-datapackage check {so.Example.flora.path}
+!cdp check {so.Example.flora.path}
 ```
 :::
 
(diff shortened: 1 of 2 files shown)


docs/guide/cli.qmd
```diff
@@ -24,7 +24,7 @@ check-datapackage
 
 ```{python}
 #| echo: false
-!uv run check-datapackage
+!cdp
```
joelostblom, line 27: The direct command `!cdp` works because the whole docs build is launched with `
uv run quarto render`, so the Quarto/Jupyter process inherits the project venv environment, and `.venv/bin` is on PATH. In that case, check-datapackage resolves to the console script from the current project environment. Having just `uv run cdp` would mess with the temporary project dir. It would be possible to run `!uv --project "$QUARTO_PROJECT_DIR" run cdp`, but that is quite noisy, so I rather treat `just build-website` as the official way to run the docs so that we can rely on that the environmental dirs are set correctly and use base `!cdp` etc in the docs themselves.

lwjohnst86 (approved):

Issue #258, opened 2025-12-18
`explain()` sometimes marks up very long string with `^`


joelostblom:
In this example:

```py
import check_datapackage as cdp
package_properties = cdp.example_package_properties()
del package_properties["resources"]
cdp.check(properties=package_properties, error=True)
```

The entire `resource` value is marked with `^`

<img width="800" height="427" alt="Image" src="https://github.com/user-attachments/assets/d8d8edff-09d9-466c-8432-602aedbf92db" />

While correct, this is not helpful and looks confusing. One possible solution would be to truncate long objects/values as `...`. This would not only be helpful for `explain()` but also avoid that we clutter the output if there is a huge input dictionary/json. So the output in this case would looks something like this (truncating everything between the first and last key):

```fortran
At resources:
|
| resources: {'name': 'woolly-dormice' ... 'licenses': [{'name': 'odc-pddl'}]}
             ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
'resources' is a required property
```

It would be interesting to investigate what other software package that use `^` for errors do in these situations.

lwjohnst86:
Also, I feel like that is the wrong output... That content shown in `resources` is actually the full package metadata. 

I also think this is an issue with how to show "required" properties. For missing required properties, it might be useful to just show "resources: <empty>" or something similar.

joelostblom:
done in #343

PR #343, opened 2026-04-29
fix: 🐛 add `<MISSING>` value instead of entire data package

joelostblom:
# Description

This is one suggestion to address #258, by adding a sentinel `<MISSING>` value to indicate missingness instead of displaying the entire datapackage's data in the output for the missing value (which was incorrect).

Before:
image

After:
image

Needs a thorough review.

## Checklist

- [x] Ran `just run-all`


lwjohnst86 (changes requested):
Oh nice 😁

lwjohnst86 (approved):

PR #345, opened 2026-04-29
fix: 🐛 show parent to indicate nesting correctly in error

### LINKED ISSUE #344: Path displayed in error could benefit from showing parent level
When putting up #343 I realized that the hierarchy suggested in the error message is a bit off. Take this example:

Image

To me, this suggest that there is an issue at the key/property `resources` which has another property called `resources` under it. I think it would be better if it said `At top level:` and then the rest kept the same.

Likewise for this example, it would be more accurate if the first line read `At resources[0]:`:

Image

joelostblom:
# Description

Closes #344, see motivation and current state there.

After this PR, the parent path is shown after `At` to make it clear that the mentioned properly is nested under that path:
image

---

image

---

When an index is last in the path, the last key is moved down together with the index:
image

Needs a thorough review.

## Checklist

- [x] Ran `just run-all`

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -17,6 +17,12 @@ individual release will not have many changes within it. Below is a list
 of the releases we've made so far, along with what was changed within
 each release.
 
+## 0.33.1 (2026-04-29)
+
+### Fix
+
+- 🐛 add `<MISSING>` value instead of entire data package (#343)
+
 ## 0.33.0 (2026-04-16)
 
 ### Feat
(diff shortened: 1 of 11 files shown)


lwjohnst86 (dismissed):
Very neat 🌮 🎉

lwjohnst86:
@joelostblom just resolve the conflicts and I will merge :+1:

joelostblom:
Resolved!

lwjohnst86 (approved):
:tada:

PR #347, opened 2026-04-30
docs: 📝 improve flow and narrative of config guide

joelostblom:
# Description

I think this is easier to follow and avoids repeating headings.

Needs a thorough review.

## Checklist

- [x] Ran `just run-all`


docs/guide/config-file.qmd
```diff
@@ -28,91 +26,105 @@ Settings provided via the CLI or Python interface always take priority
...
 
-### Configuration file syntax
+## Configuration file syntax
 
 Both configuration files use TOML syntax. The examples below start with
-`.cdp.toml`, then show the `pyproject.toml` variant.
+`.cdp.toml`, then show the `pyproject.toml` variant. All the examples
+below assume that [Soil's](https://soil.seedcase-project.org) built-in
+[`simple` Data Package](https://soil.seedcase-project.org/docs/reference/example)
+exist in the current directory as `datapackage.json`, which you can
```
martonvago:
```suggestion
exists in the current directory as `datapackage.json`, which you can
```

docs/guide/config-file.qmd
```diff
@@ -28,91 +26,105 @@ Settings provided via the CLI or Python interface always take priority
...
 
 try:
-    so.format_output_for_docs(["cdp", "check", str(so.Example.simple.path)])
+    so.format_output_for_docs(["cdp", "check"])
 finally:
     config_path.unlink(missing_ok=True)
 ```
 
-This fails because strict mode requires the `id` property.
+This is the same as if we would have run
```
martonvago:
```suggestion
This is the same as if we had run
```

docs/guide/config-file.qmd
```diff
@@ -28,91 +26,105 @@ Settings provided via the CLI or Python interface always take priority
...
+`check-datapackage check --strict` before creating the configuration
+file.
+
+If you prefer to configure `check-datapackage` via `pyproject.toml`, the
+syntax is almost the same as above; the only difference is that you need
+to include a heading to indicate where in the file `check-datapackage`
+can find the settings. Open up your `pyproject.toml` (or create one in
+the root of your project) and add a new table header
+[tool.check-datapackage]. Then put the same configuration settings as
```
martonvago:
```suggestion
`[tool.check-datapackage]`. Then put the same configuration settings as
```

docs/guide/config-file.qmd
```diff
@@ -28,91 +26,105 @@ Settings provided via the CLI or Python interface always take priority
...
+``` {.toml filename="pyproject.toml"}
+[tool.check-datapackage]
+strict = true
+```
+
+### Exclusions
 
-To exclude only that issue, add an exclusion to the `.cdp.toml` file that matches both the
-JSONPath and issue type:
+Above, we saw above that the simple example data package fails strict
```
martonvago:
```suggestion
Above, we saw that the simple example data package fails strict
```

martonvago (changes requested):
Very nice, just some typos ⭐

docs/guide/config-file.qmd
```diff
@@ -10,8 +10,6 @@ For command-line usage, see the [CLI guide](/docs/guide/cli.qmd). For
 the Python `Config` class used with `check()`, see
 [Configuring the checks](/docs/guide/config.qmd).
 
-## Configuration files
-
 A configuration file lets you set preferences once, so you don't have to
 repeat them every time you run `check-datapackage check`.
 `check-datapackage` supports two configuration files: `.cdp.toml`
```
signekb:
```suggestion
`check-datapackage` supports two configuration files: `.cdp.toml` and
```

docs/guide/config-file.qmd
```diff
@@ -28,91 +26,105 @@ Settings provided via the CLI or Python interface always take priority
...
-### Configuration file syntax
+## Configuration file syntax
 
 Both configuration files use TOML syntax. The examples below start with
-`.cdp.toml`, then show the `pyproject.toml` variant.
+`.cdp.toml`, then show the `pyproject.toml` variant. All the examples
+below assume that [Soil's](https://soil.seedcase-project.org) built-in
+[`simple` Data Package](https://soil.seedcase-project.org/docs/reference/example)
+exist in the current directory as `datapackage.json`, which you can
+accomplish by running this code snippet:
```
signekb:
```suggestion
below assume that a `datapackage.json` file exists in the current directory. Here, we use the [`simple` Data Package](https://soil.seedcase-project.org/docs/reference/example) example from our package [Soil](https://soil.seedcase-project.org). You can write the same content to `datapackage.json` in your current directory by running the following code snippet:
```
Maybe? :)

docs/guide/config-file.qmd
```diff
@@ -28,91 +26,105 @@ Settings provided via the CLI or Python interface always take priority
...
 
-First, enable strict mode without any exclusions:
+Now, create a `.cdp.toml` file and enable strict checking:
 
 ``` {.toml filename=".cdp.toml"}
 strict = true
 ```
 
-Run the check:
+Now, when you run `check-datapackage check`, strict mode is enabled and
```
signekb:
```suggestion
When you run `check-datapackage check` again, strict mode is enabled and
```
To avoid two sentences in a row starting with "now".

docs/guide/config-file.qmd
```diff
@@ -28,91 +26,105 @@ Settings provided via the CLI or Python interface always take priority
...
+#| echo: false
+import seedcase_soil as so
 
-``` {.bash filename="Terminal"}
-simple_example="$(python -c 'from seedcase_soil import Example; print(Example.simple.path)')"
+so.format_output_for_docs(["cdp", "check"])
 ```
 
-First, enable strict mode without any exclusions:
+Now, create a `.cdp.toml` file and enable strict checking:
```
signekb, line 81: Add what "strict" means here so it's a bit more beginner-friendly?

docs/guide/config-file.qmd
```diff
@@ -28,91 +26,105 @@ Settings provided via the CLI or Python interface always take priority
...
 
 ``` {.toml filename=".cdp.toml"}
 strict = true
 ```
 
-Run the check:
+Now, when you run `check-datapackage check`, strict mode is enabled and
+the example package does not pass all the stricter checks:
```
signekb:
```suggestion
the example package does not pass the stricter checks because the `id` property is missing:
```

docs/guide/config-file.qmd
```diff
@@ -28,91 +26,105 @@ Settings provided via the CLI or Python interface always take priority
...
+`check-datapackage check --strict` before creating the configuration
+file.
+
+If you prefer to configure `check-datapackage` via `pyproject.toml`, the
+syntax is almost the same as above; the only difference is that you need
+to include a heading to indicate where in the file `check-datapackage`
+can find the settings. Open up your `pyproject.toml` (or create one in
+the root of your project) and add a new table header
+[tool.check-datapackage]. Then put the same configuration settings as
+you used above in `.cdp.toml` below this table header.
```
signekb: Alternatively, a bit simpler:
```suggestion
can find the settings. Open up your `pyproject.toml` (or create one in
the root of your project) and add a new table header 
`[tool.check-datapackage]` with the same configuration settings as
you used above:
```

docs/guide/config-file.qmd
```diff
@@ -28,91 +26,105 @@ Settings provided via the CLI or Python interface always take priority
...
+strict = true
+```
+
+### Exclusions
 
-To exclude only that issue, add an exclusion to the `.cdp.toml` file that matches both the
-JSONPath and issue type:
+Above, we saw above that the simple example data package fails strict
+checking because the required `id` property is missing. To exclude only
+that issue, add an exclusion to the `.cdp.toml` file that matches both
```
signekb:
```suggestion
checking because the required `id` property is missing. To exclude
that issue from the checks, we can add an exclusion to the `.cdp.toml` file that matches both
```
Maybe? :)

docs/guide/config-file.qmd
```diff
@@ -138,10 +150,17 @@ type = "required"
...
 
 try:
-    so.format_output_for_docs(["cdp", "check", str(so.Example.simple.path)])
+    so.format_output_for_docs(["cdp", "check"])
 finally:
     config_path.unlink(missing_ok=True)
 ```
 
```
signekb: Should we include an exclusion example in the `pyproject.toml` as well?

signekb (changes requested):
Very nice 👌 Some suggestions:

docs/guide/config-file.qmd
joelostblom: It is explained under "Configurable settings" above. Do you think that is enough or should we repeat it somewhere here too?

docs/guide/config-file.qmd
joelostblom: Added!

joelostblom (commented):
Thanks for the reviews, all updated and ready for another look!

martonvago (approved):
I love it!

signekb (approved):

lwjohnst86 (approved):

PR #348, opened 2026-04-30
feat: ✨ connect required checks extensions from the config file

### LINKED ISSUE #346: Config for extensions in CLI

joelostblom:
# Description

Closes #346 (unless we want to add custom checks in a follow up PR, more details in the issue)

Needs a thorough review.

## Checklist

- [x] Ran `just run-all`


docs/guide/config-file.qmd
```diff
@@ -149,3 +152,57 @@ finally:
 
 If you want to add different exclusions, you can keep adding more
 `[[exclusions]]` below this one.
+
+### Extensions
+
+#### Required checks
+
+You can also add required checks in your configuration file. For require
```
martonvago:
```suggestion
You can also add required checks in your configuration file. To require
```

tests/test_cli.py
```diff
@@ -134,6 +135,80 @@ def test_check_passes_exclusions_from_config_to_check(
...
+                'message = "Description is required."',
+            ]
+        )
+    )
+
+    fake_source = object()
+    mock_parse_source.return_value = fake_source
+    mock_read_properties.return_value = {"name": "test-package"}
+    mock_check.return_value = []
+
```
martonvago:
```suggestion
```
I don't think you need these because mocked functions return`MagicMock` objects by default and these serve as valid inputs to other mocked functions.

docs/guide/config-file.qmd
```diff
@@ -149,3 +152,57 @@ finally:
...
+```
+
+You can add more required checks by adding more
+`[[extensions.required_checks]]` tables.
+
+#### Custom checks
+
+Custom checks require a Python function for their `check` argument, so
+define [`CustomCheck`](/docs/reference/CustomCheck.qmd) extensions in
+Python rather than in a TOML configuration file.
```
martonvago, line 232: I like this solution! Then should we raise an error if they add `[[extensions.custom_checks]]` in the config file? Because now it tries to instantiate `CustomCheck` and inevitably fails.

martonvago (changes requested):
Very nice, just small comments. ☺️

docs/guide/config-file.qmd
```diff
@@ -149,3 +152,57 @@ finally:
 
 If you want to add different exclusions, you can keep adding more
 `[[exclusions]]` below this one.
+
+### Extensions
+
+#### Required checks
+
+You can also add required checks in your configuration file. To require
+the `contributors` property, add a table titled
```
```suggestion
the `contributors` property, add use the TOML header
```

docs/guide/config-file.qmd
```diff
@@ -149,3 +152,57 @@ finally:
...
+``` {.toml filename=".cdp.toml"}
+[[extensions.required_checks]]
+jsonpath="$.contributors"
+message="The 'contributors' field is required in the Data Package properties."
+```
+
+Run the check:
+
+``` {.bash filename="Terminal"}
+check-datapackage check "$simple_example"
```
lwjohnst86: Not sure how I missed this, but this code chunk doesn't run, so using `$simple_example` is a bit confusing here. Was that the intent? You could the write the example to a file in the local folder (e.g. `_example/datapackage.json`) and use that throughout instead.

lwjohnst86 (changes requested):

docs/guide/config-file.qmd
joelostblom: This was actually already updated in the exclusions PR to use downloaded data, but not yet changed here; good that you noticed!

docs/guide/config-file.qmd
joelostblom: Agree with this. Unfortunately it ended up being a bit annoying because just adding a check in a naive way as per 92a5aad creates and ugly error message:

```
❯ cdp check                                                                                          
╭─ Error ────────────────────────────────────────────────────────────────────────────────────────────╮
│ 1 validation error for Extensions                                                                  │
│ custom_checks                                                                                      │
│   Custom checks cannot be configured in TOML because `check` must be a Python callable. Define     │
│ CustomCheck extensions in Python instead. [type=custom_checks_config, input_value=[{'type':        │
│ 'only-mit', 'js...eck': "value == 'mit'"}], input_type=list]                                       │
╰────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

The current implementation in 169b3a9  formats this error to be more readable at the expense of more complex logic for all errors just to catch this one:

```
❯ cdp check                                                                                          
╭─ Error ────────────────────────────────────────────────────────────────────────────────────────────╮
│ Custom checks cannot be configured in TOML because `check` must be a Python callable. Define       │
│ CustomCheck extensions in Python instead.                                                          │
╰────────────────────────────────────────────────────────────────────────────────────────────────────╯
```

I was initially trying to use Pydantic's hide_input_in_error=True but it seems like it is not compatible with models that are wrapped in Annotated. I considered pursuing this and fixing upstream, but I think there might still be parts of the error that is a bit opaque (like the type still being included), so I went with the custom solution in the end (but open to alternatives).

tests/test_cli.py
joelostblom: Thanks! I changed it to still check that the functions were called once, but not return any values.

docs/guide/config-file.qmd
martonvago: Hmm, I see!
The only other thought I had was setting something like:
```python
custom_checks: Annotated[
    list[CustomCheck],
    Parameter(parse=False, show=False),
] = []
```
in `Extensions`.

This gives the error:

Image

Which is not as precise as yours but maybe not horrible (but I have lower standards haha).

joelostblom: Ah interesting idea! But yeah, I think the error message probably isn't helpful enough... I'm thinking it would be confusing that is just says that `[extensions][custom_checks]` is an unknown option, since it is an option that we actually support, just not through the config. Pointing users to how to specify that option would be more helpful (although I agree that the smaller change required is attractive =p )

martonvago: Sorry for the delay, I've been thinking about this a bit because it is a problem unique to the CLI, so it would be nice if it could be solved without touching the classes that support both the CLI and the Python package. The current solution relies on detecting if the class was instantiated from the CLI, and that is information that we know in advance and shouldn't really need to recover at a later point. And, ideally, in the "generalist" classes we shouldn't need to care about how the class was created.

Some ideas:
1. Instead of having `extensions` as a CLI param, have just `required_checks`. Downsides are the diverging structures between the interfaces and the autogenerated error messages.
2. Define a separate class for the `extensions` CLI input and handle `custom_checks` there. E.g.:
```python
class ExtensionsCLI:
    def __init__(
        self,
        required_checks: list[RequiredCheck] | None = None,
        custom_checks: Any = None,
    ) -> None:
        if custom_checks is not None:
            raise ValueError(CUSTOM_CHECKS_CONFIG_ERROR)
        self.required_checks = required_checks or []
```
And later instantiate `Extensions` from this.
This would keep a separation between CLI-related logic and general logic.

3. Maybe we could do # 2 above with cyclopts [converters](https://cyclopts.readthedocs.io/en/latest/parameters.html#converters), but this might be a bit fiddly because converters don't work with Pydantic models.

What do you think?

joelostblom: > because it is a problem unique to the CLI, so it would be nice if it could be solved without touching the classes that support both the CLI and the Python package

Great point; I agree with that.

I changed to your second suggestion in 11e9714. I think this also makes the code easier to read; great idea!

martonvago (dismissed):
Yay!! 💪 🧑‍🌾 💪

src/check_datapackage/cli.py
lwjohnst86:
```suggestion
from typing import Annotated, Any, Optional
```

src/check_datapackage/cli.py
```diff
@@ -15,6 +15,27 @@
...
+    "a Python callable. Define CustomCheck extensions in Python instead."
+)
+
+
+class ExtensionsCli:
+    """Extensions that can be represented in a TOML config file."""
+
+    def __init__(
+        self,
+        required_checks: list[RequiredCheck] | None = None,
```
lwjohnst86:
```suggestion
        required_checks: Optional[list[RequiredCheck]] = None,
```

lwjohnst86 (changes requested):
Nice, only some small changes, plus the uv.lock file needs to be updated to resolve the conflict.

lwjohnst86 (approved):

PR #392, opened 2026-08-13
chore: 🔧 update configuration files from template

signekb:
# Description

Needs a quick review.


lwjohnst86 (changes requested):
This really should be two or three PRs (e.g. release process updates, devex file updates, and build check updates). Could you split this and #393 up into atomic PRs?

lwjohnst86 (approved):

PR #393, opened 2026-08-13
ci: 👷 update workflows from template

signekb:
# Description

Needs a quick review.


lwjohnst86 (changes requested):
This should be two PRs (release process with its config files and general checks with their config files (e.g. typos)). Can you actually split up this and #392 into atomic PRs? The config files in #392 for release should be with the release files here as their own PR (so build passes) and the config files for e.g. typos and checks should be with their workflows in here.

.github/workflows/add-to-project.yml
```diff
@@ -14,14 +14,45 @@ on:
...
+
+      # Using this security pattern for GitHub Apps is recommended by GitHub and ensures that
+      # the token is only available for a short time and has limited permissions. Check out
+      # <https://guidebook.seedcase-project.org/operations/security> for more details.
+      - uses: actions/create-github-app-token@1b10c78c7865c340bc4f6099eb2f838309f1e8c3 # v3.1.1
+        id: app-token
+        with:
+          # TODO: Confirm that this variable is set up for this repo.
+          client-id: "${{ vars.ADD_TO_BOARD_APP_ID }}"
+          # TODO: Confirm that this secret is set up for this repo.
```
lwjohnst86:
```suggestion
          client-id: "${{ vars.ADD_TO_BOARD_APP_ID }}"
```

lwjohnst86 (approved):

